PR #23, opened 2018-11-07
[LC-43] Leader Spinning with State Machine

windies21:
## Leader Spinning with State Machine in Siever.
 * Turn to leader when next leader is self node.
 * Reset Consensus Object when start block generate timer
 * remove made_block_count value.


loopchain/channel/channel_inner_service.py
```diff
@@ -241,16 +240,14 @@ def get_tx_info(self, tx_hash):
 
         self._channel_service.state_machine.vote()
 
-        if unconfirmed_block.made_block_count >= conf.LEADER_BLOCK_CREATION_LIMIT \
-                and unconfirmed_block.block_type is BlockType.vote \
-                and unconfirmed_block.is_divided_block is False:
+        if unconfirmed_block.block_type is BlockType.vote and unconfirmed_block.is_divided_block is False:
```
yakkle:
```suggestion
        if unconfirmed_block.block_type is BlockType.vote and unconfirmed_block.is_divided_block:
```
if is_divided_block is bool value why compare with bool?

windies21: This is not a modification in this PR.

zsaladin (approved):

PR #26, opened 2018-11-15
revert old timer in subscribe(channel_service) for compatibility

laurenceyoon:
Supports backward compatibility so that citizen node can subscribe continuously even it's disconnected from previous versions of the network.

### DIFF
--- a/loopchain/channel/channel_service.py
+++ b/loopchain/channel/channel_service.py
@@ -497,20 +497,90 @@ class ChannelService:
                                                block_height=self.block_manager.get_blockchain().block_height)
 
     async def __subscribe_call_by_rest_stub(self, rs_rest_stub):
-        if conf.REST_SSL_TYPE == conf.SSLAuthType.none:
-            peer_target = ChannelProperty().rest_target
+        response = {'response_code': message_code.Response.fail,
+                    'message': message_code.get_response_msg(message_code.Response.fail)}
+
+        try:
+            if conf.REST_SSL_TYPE == conf.SSLAuthType.none:
+                peer_target = ChannelProperty().rest_target
+            else:
+                peer_target = f"https://{ChannelProperty().rest_target}"
+            response = rs_rest_stub.call(
+                "Subscribe", {
+                    'channel': ChannelProperty().name,
+                    'peer_target': peer_target
+                }
+            )
+
+        except Exception as e:
+            logging.warning(f"Due to Subscription fail to RadioStation(mother peer), "
+                            f"automatically retrying subscribe call")
+
+        if response['response_code'] == message_code.Response.success:
+            if TimerService.TIMER_KEY_SUBSCRIBE in self.__timer_service.timer_list.keys():
+                self.__timer_service.stop_timer(TimerService.TIMER_KEY_SUBSCRIBE)
+                logging.debug(f"Subscription to RadioStation(mother peer) is successful.")
+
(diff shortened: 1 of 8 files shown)


loopchain/channel/channel_service.py
```diff
@@ -497,20 +497,90 @@ def connect_to_radio_station(self, is_reconnect=False):
...
-            peer_target = ChannelProperty().rest_target
+        response = {'response_code': message_code.Response.fail,
+                    'message': message_code.get_response_msg(message_code.Response.fail)}
+
+        try:
+            if conf.REST_SSL_TYPE == conf.SSLAuthType.none:
+                peer_target = ChannelProperty().rest_target
+            else:
+                peer_target = f"https://{ChannelProperty().rest_target}"
+            response = rs_rest_stub.call(
```
zsaladin: `rs_rest_stub.call()` could be blocked.
I recommnend replacing it with `await rs_rest_stub.call_async()`

loopchain/channel/channel_service.py
```diff
@@ -497,20 +497,90 @@ def connect_to_radio_station(self, is_reconnect=False):
...
+                        target=timer_key,
+                        duration=conf.SHUTDOWN_TIMER,
+                        callback=self.shutdown_peer,
+                        callback_kwargs={"message": error}
+                    )
+                )
+
+        return response
+
+    def __check_block_height_call_to_rs_stub(self, **kwargs):
```
zsaladin: def -> async def 
Then you can use `rest_stub.call_async`

loopchain/channel/channel_service.py
```diff
@@ -497,20 +497,90 @@ def connect_to_radio_station(self, is_reconnect=False):
...
+        rs_rest_stub = kwargs.get("rs_rest_stub", None)
+        try:
+            last_block = rs_rest_stub.call("GetLastBlock")
+            if last_block['height'] <= self.__block_manager.get_blockchain().block_height:
+                return
+            else:
+                # citizen needs additional block or connected to problematic peer.
+                raise Exception
```
zsaladin: This exception does not seem useful.
Why don't you remove `try, except` and `else`

zsaladin (approved):

loopchain/channel/channel_service.py
laurenceyoon: Updated to async call

loopchain/channel/channel_service.py
laurenceyoon: updated to call_async and changed method to async.

loopchain/channel/channel_service.py
```diff
@@ -501,16 +501,79 @@ def connect_to_radio_station(self, is_reconnect=False):
...
+        if last_block['height'] <= self.__block_manager.get_blockchain().block_height:
+            return
+
+        timer_key = TimerService.TIMER_KEY_GET_LAST_BLOCK_KEEP_CITIZEN_SUBSCRIPTION
+        if timer_key in self.__timer_service.timer_list.keys():
+            util.logger.spam(f"stop timer for check_block_height_call to radiostation...")
+            self.__timer_service.stop_timer(timer_key)
+
+        timer_key = TimerService.TIMER_KEY_SUBSCRIBE
+        if timer_key not in self.__timer_service.timer_list.keys():
```
yakkle:
```suggestion
        if timer_key not in self.__timer_service.timer_list:
```
to check key in dict, don't need get dict.keys().

laurenceyoon: Updated.

yakkle (approved):

PR #35, opened 2018-12-04
RC1.23.1

windies21:
Block Versioning System integration - Release (1.23.0) bug fix
- Deploy Tx generation hash and verification hash mismatch - HashGenerator version setting error during block validation
- V2 To Address error - Unmatched address - MalformedAddress type created and handled as str (no byte processing)
- V2 Nonce value format error - Unmatched format such as "0x1", "64545" - Process with str
- V2 Value Format Error - Hexadecimal number without "0x" - For v2, all processing with str (fee included)
- V2 Extra value format error - no relevant object such as "lang" exists in params
- V3 Data value format error - Hexadecimal number without "0x" despite "DataType" being "message" - All processing with str
- Modify Citizen to receive ConfirmedBlock from mainnet and test net without websocket

### DIFF
--- a/loopchain/blockchain/blocks/__init__.py
+++ b/loopchain/blockchain/blocks/__init__.py
@@ -3,3 +3,5 @@ from .block_builder import BlockBuilder
 from .block_serializer import BlockSerializer
 from .block_verifier import BlockVerifier
 from .block_versions import BlockVersions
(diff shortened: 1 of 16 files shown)


hangaroy (approved):
Approved

PR #39, opened 2018-12-10
Fix multiple leader problem when peers try to connect at a same time

jckdotim:
# Bug

This patch fixes two initial run problems.

1. In some environments, `RestServiceRS` will not run when we execute `loop rs`. The `time.sleep (1)` fixes this situation, but I added a TODO comment to encourage you to determine the exact cause.

2. Fixed an issue that occurs when multiple peers attempt to connect to the initialized loopchain at the same time. `__load_peer_manager_lock` locks total process of creating `PeerManager` and new peer information.


yakkle:

yakkle:
ok to test

yakkle:
test this please

PR #65, opened 2019-01-28
[LC-122] Leader Complain without RS

windies21:
* Leader Complain without RS
  - Add Leader Complain Timer
  - Add is_complain flag "0.2" block header
  - Update leader complain scenario
  - Add Epoch Class


loopchain/channel/channel_service.py
```diff
@@ -457,6 +461,17 @@ def connect_to_radio_station(self, is_reconnect=False):
...
+    def __subscribe_to_peer_list(self):
+        peer_object = self.peer_manager.get_peer(ChannelProperty().peer_id)
+        peer_request = loopchain_pb2.PeerRequest(
+            channel=ChannelProperty().name,
+            peer_target=ChannelProperty().peer_target,
+            peer_id=ChannelProperty().peer_id, group_id=ChannelProperty().group_id,
+            node_type=ChannelProperty().node_type,
+            peer_order=peer_object.order
+        )
+        self.__broadcast_scheduler.schedule_broadcast("Subscribe", peer_request)
```
hangaroy: why don't you use BroadcastCommand.SUBSCRIBE.

hangaroy: this is my mistake.

zsaladin (approved):

loopchain/baseservice/timer_service.py
```diff
@@ -129,6 +130,19 @@ def add_timer(self, key, timer):
         if timer.is_run_at_start:
             self.restart_timer(key)
 
+    def add_timer_convenient(self, timer_key, duration, is_repeat=False, callback=None, callback_kwargs=None):
```
zsaladin: It seems a shortcut of `add_timer`. But it has another logic that 
```python
if timer_key not in self.__timer_list
```

I think the method should create new timer and add that timer without 'if statement'

loopchain/blockchain/blockchain.py
```diff
@@ -261,6 +261,12 @@ def __add_block(self, block: Block, vote: Vote = None):
                     'block_height': self.__block_height
                 }})
 
+            # stop leader complain timer
```
zsaladin: It makes coupling between blockchain and consensus algorithm implementations. I cannot help it now, but I think it would be better to remove it later.

loopchain/blockchain/epoch.py
```diff
@@ -0,0 +1,62 @@
...
+        self.leader_id = leader_id
+
+        # TODO using Epoch in BlockManager instead using candidate_blocks directly.
+        # But now! only collect leader complain votes.
+        self.__candidate_blocks = None
+        self.__complain_vote = Vote(Epoch.COMPLAIN_VOTE_HASH, ObjectManager().channel_service.peer_manager)
+
+    @staticmethod
+    def new_epoch(height: int, leader_id=None):
+        if leader_id is None and ObjectManager().channel_service.block_manager.epoch.leader_id:
```
zsaladin: 'and ObjectManager().channel_service.block_manager.epoch.leader_id' is not needed.
It can be modified.
```python
leader_id = leader_id or ObjectManager().channel_service.block_manager.epoch.leader_id
```

loopchain/channel/channel_inner_service.py
```diff
@@ -402,6 +392,9 @@ def block_height_sync(self):
 
     @message_queue_task(type_=MessageQueueType.Worker)
     def add_audience(self, peer_target) -> None:
+        peer = self._channel_service.peer_manager.get_peer_by_target(peer_target)
+        if not peer:
+            util.logger.debug(f"There is no peer peer_target({peer_target})")
```
zsaladin: In this case there is not follow-up. Just logging

PR #73, opened 2019-02-08
[LC-165] Use time.monotonic function instead of time.clock_gettime function

hangaroy:
- There is no time.clock_gettime function some system even if it is unix system
- time.monotonic is system-wide function after python 3.5


zsaladin (approved):

zsaladin:
Can you add ticket number to the title?

hangaroy:
> Can you add ticket number to the title?

Added

PR #75, opened 2019-02-11
[LC-130] update README & loading key as keystore file

laurenceyoon:
- generate key from tbears and apply it.
- separate run Citizen Node manual from existing README.
- delete unused scripts
- add init_key_by_channel


sendtx_testnet.json
```diff
@@ -2,10 +2,7 @@
   "jsonrpc": "2.0",
   "method": "icx_sendTransaction",
   "params": {
-    "version": "0x3",
```
zsaladin: Is there any reason to remove them?

laurenceyoon: Only the required params are left in the json file which is for sendtx through tbears command.

laurenceyoon: This file was deleted since the icx transfer method was changed.

windies21 (approved):

PR #90, opened 2019-02-13
RC2.0.3

laurenceyoon:
LC-77 - fix bug in v1 multichannel request
LC-173 - add channel filter when loading channels json file by peer
LC-168 - prevent leader complain timer when peer is leader


yakkle (approved):
👌

PR #99, opened 2019-02-18
[LC-174] update conf according to iconrpcserver update

laurenceyoon:
- apply `redirectProtocol` as json conf as well as env variable.
- separate the conf loading from dev_mode and none dev_mode (run as citizen node to testnet / mainnet)


laurenceyoon:
This pull request should be deferred until iconrpcserver is released to master.

loopchain/launcher.py
```diff
@@ -108,28 +108,35 @@ def start_as_rest_server(args):
...
+    rpcserver_conf: IconConfig = IconConfig("", default_rpcserver_config)
     rpcserver_conf.load()
     rpcserver_conf.update_conf(additional_conf)
+    if dev_mode:
+        additional_conf.update({
+            "channel": channel,
+            "port": api_port,
+            "amqpKey": amqp_key
+        })
+        rpcserver_conf.update_conf(additional_conf)
```
yakkle: It would be better `additional_conf.update` place on above create instance of `IconConfig` line.
then, `rpcserver_conf.update_conf` will be called once when dev_mode.

laurenceyoon: As you mentioned, I removed the part that calls update_conf (additional_conf) twice. Thank you.

windies21 (approved):

PR #98, opened 2019-02-18
[LC-141] Add `BloomFilter`

zsaladin:
# BloomFilter
- Merge transaction receipts logsBloom

PR #100, opened 2019-02-18
[LC-185] Add `BlockProver`

zsaladin:
## BlockProve
- `get_proof` - It will return proof of target hash. The proof is a list consisting of leaves of merkle tree.
- `get_proof_root` - It will return root of merkle tree.
- `prove` - It will return True if the proof as a parameter is valid for target hash and root hash.


zsaladin:
Note that it must be merge after merging #98 LC-141.

PR #108, opened 2019-02-21
[LC-198] Remove pickle comments

metalg0su:
- remove all comments: Changing the import location will cause a pickle error."
 - optimized import orders


yakkle:

yakkle:
ok to test

PR #116, opened 2019-02-22
[LC-204] Remove block version 0.2

windies21:
Do leader complain without block version up.


loopchain/blockchain/epoch.py
```diff
@@ -118,12 +118,14 @@ def __add_tx_to_block(self, block_builder):
...
         # self._check_unconfirmed_block(
-        block_height = self.__blockchain.last_block.header.height + 1
+        last_block = self.__blockchain.last_unconfirmed_block or self.__blockchain.last_block
+        block_height = last_block.header.height + 1
         block_version = self.__blockchain.block_versioner.get_version(block_height)
         block_builder = BlockBuilder.new(block_version, self.__blockchain.tx_versioner)
         if self.complain_result():
             block_builder.complained = self.complain_result()
         else:
+            block_builder.complained = False
```
zsaladin: BlockBuilder 0.1a does not have `complained` attribute.

loopchain/blockchain/epoch.py
```diff
@@ -116,14 +116,13 @@ def __add_tx_to_block(self, block_builder):
                 block_builder.transactions[tx.hash] = tx
                 block_tx_size += tx.size(tx_versioner)
 
-    def makeup_block(self):
+    def makeup_block(self, complained: str or None):
```
yakkle: I thinks `complained` is bool type. right?
And it's not  correct annotation syntax.
https://stackoverflow.com/a/33945518

yakkle (approved):

PR #124, opened 2019-02-28
[LC-216] fix bug in blocksync by leader

laurenceyoon:
- Fixed a bug that peer cannot sync in BlockSync state when the peer sets itself as a BlockGenerator.
- update Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,14 +51,26 @@ check:
 test:
 	@python3 -m unittest discover testcase/unittest/ -p "test_*.py" || exit -1
 
-# clean-mq clean-build clean-pyc
-clean: clean-mq clean-build clean-pyc
+# Clean all: clean-process clean-mq clean-pyc clean-db clean-log
+clean: clean-process clean-mq clean-pyc clean-db clean-log
+
+linux-clean: clean-process clean-mq-linux clean-pyc clean-db clean-log
+
+clean-process:
+	pkill -f loop || true
+	pkill -f gunicorn || true
 
 clean-mq:
 	@echo "Cleaning up RabbitMQ..."
 	@rabbitmqctl stop_app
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -51,15 +51,27 @@ check:
 test:
 	@python3 -m unittest discover testcase/unittest/ -p "test_*.py" || exit -1
 
-# clean-mq clean-build clean-pyc
-clean: clean-mq clean-build clean-pyc
+# Clean all: clean-process clean-mq clean-pyc clean-db clean-log
+clean: clean-process clean-mq clean-pyc clean-db clean-log
+
+linux-clean: clean-process clean-mq-linux clean-pyc clean-db clean-log
```
yakkle: you can check OS platform by 'uname'. check [this](https://stackoverflow.com/a/24563202).

loopchain/channel/channel_service.py
```diff
@@ -593,57 +584,9 @@ def __ready_to_height_sync(self):
...
-
-                self.show_peers()
-
-            if is_delay_announce_new_leader and ChannelProperty().node_type == conf.NodeType.CommunityNode:
-                self.peer_manager.announce_new_leader(
-                    peer_old_leader.peer_id,
-                    peer_leader.peer_id,
-                    self_peer_id=ChannelProperty().peer_id)
+        _, future = self.block_manager.block_height_sync()
+        await future
```
hangaroy: I think it can't wait concurrent.futures.Future with await

await (asyncio.Future) loop.run_in_executor(...)

Makefile
laurenceyoon: Applied. Thank you very much for the reference.

windies21 (approved):

PR #141, opened 2019-03-12
[LC-256] Doesn't stop leader complain timer when node receive an unconfirmed block again

hangaroy:
when node receives unconfirmed block that equals previous unconfirmed block, node just vote again.


yakkle:
retest this please

yakkle:
retest this please

PR #146, opened 2019-03-20
RC2.0.11

windies21:
LC-246 - fix subscribe error handling for citizen node RELEASE CANDIDATE
LC-248 - Log analysis for Citizen Sync fail issue in Mainnet RELEASE CANDIDATE


yakkle (approved):

Makefile
```diff
@@ -27,16 +27,13 @@ all: install generate
...
-	pip3 install git+https://github.com/icon-project/icon-service.git@master
-	pip3 install git+https://github.com/icon-project/icon-commons.git@master
-	pip3 install git+https://github.com/icon-project/icon-rpc-server.git@master
-	pip3 install tbears
+install:
+	pip3 install -U git+https://github.com/icon-project/icon-service.git@master
+	pip3 install -U git+https://github.com/icon-project/icon-commons.git@master
+	pip3 install -U git+https://github.com/icon-project/icon-rpc-server.git@master
+	pip3 install -U tbears
 	pip3 install -e .
```
zsaladin: Doesn't it need '-U' option?

hangaroy (approved):

Makefile
laurenceyoon: The -U option allows you to install the library regardless of whether the version or some of the files of the library has changed or not. (In case the master version has not changed and some files have been modified)

loopchain/channel/channel_service.py
```diff
@@ -521,16 +521,16 @@ def __unsubscribe_call_by_rest_stub(self):
...
-            return
-
-        # RS peer Connection Error
-        self.__unsubscribe_call_by_rest_stub()
-        self.stop_check_last_block_rs_timer()
-        if self.__state_machine.state != "SubscribeNetwork":
-            logging.error(f"changing state to subscribe network...")
-            self.__state_machine.subscribe_network()
+        try:
+            self.__radio_station_stub.call("GetLastBlock")
```
zsaladin: `last_block` condition statement has disappeared.

laurenceyoon: The function to check the last block is only used to check that the RS peer is alive. The comparison of the `last_block` is excluded because it is also done by annouce_confirmed_block as well.

loopchain/channel/channel_service.py
```diff
@@ -496,23 +492,42 @@ def _handle_exception(future: asyncio.Future):
...
-            logging.debug(f"Subscription to RadioStation(mother peer) is successful.")
+            logging.debug(f"Subscription to RS peer({ChannelProperty().radio_station_target}) is successful.")
             event.set()
             self.start_check_last_block_rs_timer()
 
-    def __check_last_block_to_rs(self):
-        last_block = self.__radio_station_stub.call_async("GetLastBlock")
-        if last_block['height'] <= self.__block_manager.get_blockchain().block_height:
+    def __unsubscribe_call_by_rest_stub(self):
+        if conf.REST_SSL_TYPE == conf.SSLAuthType.none:
```
zsaladin: This conf is this node's option not rs's option.
So in my opinion it is dangerous to determine rs's option based on my option.

loopchain/channel/channel_service.py
zsaladin: Then you mean this change does not affect original behaviour, right?

laurenceyoon: Yep:)

PR #152, opened 2019-03-26
[LC-270] refactoring announce_unconfirmed_block to verify block_info when confirm prev block.

windies21:
* Call state machine trigger 'vote' with unconfirmed block param.

* remove __unconfirmedBlockQueue

* Send failed vote when AddUnconfirmedBlock exception raised.

* set TIMEOUT_FOR_LEADER_COMPLAIN to 60 (1 min.)

### DIFF
--- a/loopchain/peer/block_manager.py
+++ b/loopchain/peer/block_manager.py
@@ -62,7 +62,6 @@ class BlockManager:
         )
         self.__txQueue = AgingCache(max_age_seconds=conf.MAX_TX_QUEUE_AGING_SECONDS,
                                     default_item_status=TransactionStatusInQueue.normal)
-        self.__unconfirmedBlockQueue = queue.Queue()
         self.__blockchain = BlockChain(self.__level_db, channel_name)
         self.__peer_type = None
         self.__consensus = None
@@ -276,8 +275,6 @@ class BlockManager:
             return False
 
         self.epoch.set_epoch_leader(unconfirmed_block.header.next_leader.hex_hx())
-
-        self.__unconfirmedBlockQueue.put(unconfirmed_block)
 
         return True
 
@@ -576,7 +573,6 @@ class BlockManager:
         if self.__blockchain.last_unconfirmed_block is not None:
             self.candidate_blocks.remove_block(self.__blockchain.last_unconfirmed_block.header.hash)
         self.__blockchain.last_unconfirmed_block = None
-        self.__unconfirmedBlockQueue = queue.Queue()
 
         my_height = self.__current_block_height()
         logging.debug(f"in __block_height_sync max_height({max_height}), my_height({my_height})")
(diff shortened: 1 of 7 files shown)


loopchain/peer/block_manager.py
```diff
@@ -765,44 +764,55 @@ def vote_unconfirmed_block(self, block_hash, is_validated):
...
-
-        reps = self.__channel_service.get_rep_ids()
-        exception = None
+        _exception = None
         try:
+            if self.add_unconfirmed_block(unconfirmed_block):
+                if self.__channel_service.peer_manager.get_leader_id(conf.ALL_GROUP_ID) != \
+                        unconfirmed_block.header.next_leader.hex_hx():
+                    util.logger.debug(f"reset leader to ({unconfirmed_block.header.next_leader.hex_hx()})")
+                    await self.__channel_service.reset_leader(unconfirmed_block.header.next_leader.hex_hx())
```
zsaladin: Does `reset_leader` have to happen during voting?
In my opinion the voting behaviour itself should not change state of consensus.

loopchain/peer/block_manager.py
```diff
@@ -765,44 +764,55 @@ def vote_unconfirmed_block(self, block_hash, is_validated):
...
 
-        logging.info("PeerService received unconfirmed block: " + unconfirmed_block.header.hash.hex())
-
-        block_version = self.__blockchain.block_versioner.get_version(unconfirmed_block.header.height)
-        block_verifier = BlockVerifier.new(block_version, self.__blockchain.tx_versioner)
-        block_verifier.invoke_func = self.__channel_service.score_invoke
-
-        reps = self.__channel_service.get_rep_ids()
-        exception = None
+        _exception = None
```
zsaladin: How about using `exc = None`
Note that PEP8 recommends using trailing under score on conflicting with reserved keyword.
Of course it is not the case of reserved keyword.

loopchain/peer/block_manager.py
```diff
@@ -741,44 +740,51 @@ def vote_unconfirmed_block(self, block_hash, is_validated):
...
-            self.vote_unconfirmed_block(unconfirmed_block.header.hash, exception is None)
+            self.vote_unconfirmed_block(unconfirmed_block.header.hash, exc is None)
+
+    async def vote_as_peer(self, unconfirmed_block: Block):
+        """Vote to AnnounceUnconfirmedBlock
+        """
+        util.logger.info(f"in vote_as_peer unconfirmed_block({unconfirmed_block.header.hash.hex()})")
+
+        await self._vote(unconfirmed_block)
+        await self.__channel_service.reset_leader(unconfirmed_block.header.next_leader.hex_hx())
```
zsaladin: `ResetLeader` has to be executed only if the node does True vote.

PR #159, opened 2019-04-03
[LC-287] Fixed the time delay until tx was confirmed

windies21:
Fixed the time delay until tx was confirmed

### DIFF
--- a/loopchain/peer/consensus_siever.py
+++ b/loopchain/peer/consensus_siever.py
@@ -13,13 +13,14 @@
 # limitations under the License.
 """A consensus class based on the Siever algorithm for the loopchain"""
 
+import asyncio
 import threading
 from functools import partial
 
 import loopchain.utils as util
 from loopchain import configure as conf
 from loopchain.baseservice import ObjectManager, TimerService, SlotTimer, Timer
-from loopchain.blockchain import ExternalAddress, Hash32, Vote, Block
+from loopchain.blockchain import ExternalAddress, Vote, Block
 from loopchain.channel.channel_property import ChannelProperty
 from loopchain.peer.consensus_base import ConsensusBase
 
@@ -29,6 +30,9 @@ class ConsensusSiever(ConsensusBase):
         super().__init__(block_manager)
         self.__block_generation_timer = None
         self.__lock = threading.Lock()
+
+        self._loop: asyncio.BaseEventLoop = None
+        self._vote_queue: asyncio.Queue = None
 
     def start_timer(self, timer_service):
         self.__block_generation_timer = SlotTimer(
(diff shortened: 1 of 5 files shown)


loopchain/peer/consensus_siever.py
```diff
@@ -135,36 +157,34 @@ def __build_candidate_block(self, block_builder, next_leader, vote_result):
...
-    #             vote_result = await asyncio.wait_for(self._vote_queue.get(), timeout=timeout)
-    #             if vote_result is None:  # sentinel
-    #                 return False
-    #
-    #         except asyncio.TimeoutError:
-    #             logging.warning("Timed Out Block not confirmed duration: " +
-    #                             str(util.diff_in_seconds(candidate_block.header.timestamp)))
-    #             return False
+                if not conf.ALLOW_MAKE_EMPTY_BLOCK:
+                    self.__block_generation_timer.call_ignore_slot()
```
zsaladin: How about changing the method name?
For example `call_instantly`, `call_immediately`

loopchain/peer/consensus_siever.py
```diff
@@ -135,36 +157,34 @@ def __build_candidate_block(self, block_builder, next_leader, vote_result):
...
-    #
-    #             vote_result = await asyncio.wait_for(self._vote_queue.get(), timeout=timeout)
-    #             if vote_result is None:  # sentinel
-    #                 return False
-    #
-    #         except asyncio.TimeoutError:
-    #             logging.warning("Timed Out Block not confirmed duration: " +
-    #                             str(util.diff_in_seconds(candidate_block.header.timestamp)))
-    #             return False
+                if not conf.ALLOW_MAKE_EMPTY_BLOCK:
```
zsaladin: This configuration name seems to be not correct.

loopchain/peer/consensus_siever.py
zsaladin: Or just call `await consensus()`

loopchain/peer/consensus_siever.py
```diff
@@ -125,8 +145,10 @@ def __build_candidate_block(self, block_builder, next_leader, vote_result):
             self._blockchain.last_unconfirmed_block = candidate_block
             broadcast_func = partial(self._block_manager.broadcast_send_unconfirmed_block, candidate_block)
 
-            # TODO Temporary ignore below line for developing leader complain
             self.__start_broadcast_send_unconfirmed_block_timer(broadcast_func)
+            success = await self._wait_for_voting(candidate_block)
```
zsaladin: Other voting points are not handled(line: 126). Is it okay?

loopchain/peer/consensus_siever.py
windies21: I agree with your opinion, but the configuration name is not made by this branch. We will consider it in another branch.

PR #166, opened 2019-04-08
[LC-291] Show round in epoch for debug.

windies21:
* Add start method to block generator timer in consensus.
* Add round into epoch.
* Make new epoch after block sync. And Allow add complain when my epoch has same height with complain message. If the epoch height is lower than the message, execute block sync.

### DIFF
--- a/loopchain/baseservice/slot_timer.py
+++ b/loopchain/baseservice/slot_timer.py
@@ -24,14 +24,16 @@ class SlotTimer:
         self.__delayed = True
         self.__timer_key = timer_key
         self.__timer_service = timer_service
+        self.__duration = duration
         self.__callback = callback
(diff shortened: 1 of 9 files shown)


loopchain/baseservice/slot_timer.py
```diff
@@ -54,12 +59,14 @@ def call(self):
         util.logger.spam(f"call slot({self.__slot}) delayed({self.__delayed})")
         if self.__slot > 0:
             self.__slot -= 1
-            self.__timer_service.get_event_loop().create_task(self.__callback())
+            asyncio.get_event_loop().create_task(self.__callback())
```
hangaroy: this is not root cause. please see https://github.com/icon-project/loopchain/pull/168

PR #169, opened 2019-04-10
[LC-301] add InvalidMesssage exception in node_subscriber

laurenceyoon:


zsaladin:
Could you leave a description?

laurenceyoon:
This branch is to additionally handle an exception when the citizen node attempts to subscribe to an earlier version of the parent node.

windies21 (approved):

PR #174, opened 2019-04-25
Modify statemachine for statemachine can support multiple instance

yourhrh:
Modify statemachine for statemachine can support multiple instance

Before code method __init__  adjust to class
After __init__ adjust to instance


yakkle:

yakkle:
add to whitelist

PR #177, opened 2019-04-29
[LC-325] Edit logic to manage the duration of leader complaint.

hyerim-kim:
- The duration of leader complaint belongs to `Epoch` and increases on every round until max duration.

### DIFF
--- a/loopchain/blockchain/epoch.py
+++ b/loopchain/blockchain/epoch.py
@@ -46,6 +46,7 @@ class Epoch:
         self.__complain_vote = dict()  # complain vote dict { round : Vote }
         self.__complain_vote[self.round] = Vote(Epoch.COMPLAIN_VOTE_HASH, ObjectManager().channel_service.peer_manager)
         self.complained_result = None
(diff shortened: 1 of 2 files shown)


loopchain/blockchain/epoch.py
```diff
@@ -46,6 +46,7 @@ def __init__(self, block_manager, leader_id=None):
         self.__complain_vote = dict()  # complain vote dict { round : Vote }
         self.__complain_vote[self.round] = Vote(Epoch.COMPLAIN_VOTE_HASH, ObjectManager().channel_service.peer_manager)
         self.complained_result = None
+        self.complain_duration = conf.TIMEOUT_FOR_LEADER_COMPLAIN
```
zsaladin: How about calculating each time instead of storing value?

```python
@property
def complain_duration(self):
    return (2 ** self.round) * conf.TIMEOUT_FOR_LEADER_COMPLAIN
```

laurenceyoon (approved):

zsaladin (approved):

PR #180, opened 2019-04-30
[LC-312] Add switching role between P-Rep and Citizen

hangaroy:
- Switch role from Citizen to Rep when a block of specific height is added.
- Reset network with state machine.
- ResetNetwork -> EvaluateNetwork -> ...
- create_icx_tx() function return fail_no_permission if node is citizen.
  (RPCServer always call create_icx_tx. if fail_no_permission is returned, RPCServer redirect the tx to parent node)

### DIFF
--- a/loopchain/baseservice/node_subscriber.py
+++ b/loopchain/baseservice/node_subscriber.py
@@ -26,6 +26,7 @@ from jsonrpcserver.aio import AsyncMethods
 from websockets.exceptions import InvalidStatusCode, InvalidMessage
 
 from loopchain import configure as conf
+from loopchain import utils
 from loopchain.baseservice import ObjectManager, TimerService, Timer
 from loopchain.blockchain import BlockSerializer, BlockVerifier
 from loopchain.channel.channel_property import ChannelProperty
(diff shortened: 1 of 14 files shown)


loopchain/baseservice/node_subscriber.py
```diff
@@ -25,6 +25,7 @@
 from jsonrpcserver.aio import AsyncMethods
 from websockets.exceptions import InvalidStatusCode, InvalidMessage
 
+import loopchain.utils as util
```
windies21: I suggest this. -> from loopchain import utils

loopchain/channel/channel_statemachine.py
```diff
@@ -47,6 +47,7 @@ class ChannelStateMachine(object):
                     on_enter='_blockgenerate_on_enter', on_exit='_blockgenerate_on_exit'),
               State(name='LeaderComplain', ignore_invalid_triggers=True,
                     on_enter='_leadercomplain_on_enter', on_exit='_leadercomplain_on_exit'),
+              State(name='ResetNetwork', ignore_invalid_triggers=True),
```
windies21: Please update state machine diagram when state machine code.

metalg0su: State machine diagram updated.

windies21 (approved):

PR #183, opened 2019-05-03
[LC-330] Set peer type to PEER after block sync

hangaroy:
1. Peer is a leader and block height is less than other peers.
2. Sync block by complanning leader.
3. The peer is not changed to no leader even if the peer is not leader anymore

In Addtion,
change lock for consensus from threading.Lock to asyncio.Lock

### DIFF
--- a/loopchain/peer/peer_outer_service.py
+++ b/loopchain/peer/peer_outer_service.py
@@ -661,17 +661,6 @@ class PeerOuterService(loopchain_pb2_grpc.PeerServiceServicer):
 
         return loopchain_pb2.CommonReply(response_code=message_code.Response.success, message="success")
 
-    def AnnounceNewLeader(self, request, context):
-        if not request.channel:
-            raise Exception("peer_outer_service:AnnounceNewLeader : Channel is not defined.")
-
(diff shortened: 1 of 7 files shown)


yakkle:
please rebase with release/2.1 branch.

zsaladin (approved):

loopchain/peer/peer_outer_service.py
```diff
@@ -661,17 +661,6 @@ def VoteUnconfirmedBlock(self, request, context):
...
-        if not request.channel:
-            raise Exception("peer_outer_service:AnnounceNewLeader : Channel is not defined.")
-
-        logging.debug(f"AnnounceNewLeader({request.channel}): " + request.message)
-
-        channel_stub = StubCollection().channel_stubs[request.channel]
-        channel_stub.sync_task().reset_leader(request.new_leader_id)
-
-        return loopchain_pb2.CommonReply(response_code=message_code.Response.success, message="success")
-
```
yakkle: I think that would be better add `docstring` in announceNewLeader `deprecated` and nothing to do.
Then, refactoring with  removed radio station features.

PR #190, opened 2019-05-14
[LC-338] block generation with 2 seconds

windies21:
There are two strategy for block generation in loopchain.

(ALLOW_MAKE_EMPTY_BLOCK = False)

One is treat tx immediately and make Block also.

But if there is no tx, loopchain is wait tx without repeated empty block generation.

(ALLOW_MAKE_EMPTY_BLOCK = True)

Another is make Block repeat whether tx is or not.

Option ALLOW_MAKE_EMPTY_BLOCK can effect that strategy.

### DIFF
--- a/loopchain/blockchain/blockchain.py
+++ b/loopchain/blockchain/blockchain.py
@@ -12,13 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Block chain class with authorized blocks only"""
+
 import json
-import leveldb
 import pickle
 import threading
 import zlib
 from enum import Enum
+from typing import TYPE_CHECKING
 from typing import Union, List
+
+import leveldb
 
 import loopchain.utils as util
 from loopchain import configure as conf
@@ -30,6 +33,9 @@ from loopchain.blockchain.exception import *
 from loopchain.blockchain.score_base import *
 from loopchain.channel.channel_property import ChannelProperty
 from loopchain.utils.message_queue import StubCollection
+
+if TYPE_CHECKING:
+    from loopchain.peer import BlockManager
 
 
 class NID(Enum):
@@ -661,7 +667,8 @@ class BlockChain:
         #                    f"tx count({len(current_block.body.transactions)}), "
         #                    f"height({current_block.header.height})")
 
-        candidate_blocks = ObjectManager().channel_service.block_manager.candidate_blocks
+        block_manager: BlockManager = ObjectManager().channel_service.block_manager
+        candidate_blocks = block_manager.candidate_blocks
         with self.__confirmed_block_lock:
             logging.debug(f"BlockChain:confirm_block channel({self.__channel_name})")
 
@@ -678,7 +685,7 @@ class BlockChain:
             except KeyError:
                 if self.last_block.header.hash == current_block.header.prev_hash:
                     logging.warning(f"Already added block hash({current_block.header.prev_hash.hex()})")
-                    if current_block.header.complained:
+                    if current_block.header.complained and block_manager.epoch.complained_result:
                         util.logger.debug("reset last_unconfirmed_block by complain block")
                         self.last_unconfirmed_block = current_block
                     return None
--- a/loopchain/blockchain/blockchain.py
+++ b/loopchain/blockchain/blockchain.py
@@ -667,7 +667,7 @@ class BlockChain:
         #                    f"tx count({len(current_block.body.transactions)}), "
         #                    f"height({current_block.header.height})")
 
-        block_manager: BlockManager = ObjectManager().channel_service.block_manager
+        block_manager: 'BlockManager' = ObjectManager().channel_service.block_manager
         candidate_blocks = block_manager.candidate_blocks
         with self.__confirmed_block_lock:
             logging.debug(f"BlockChain:confirm_block channel({self.__channel_name})")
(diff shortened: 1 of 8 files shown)


loopchain/peer/consensus_siever.py
```diff
@@ -162,8 +173,7 @@ def __build_candidate_block(self, block_builder, next_leader, vote_result):
             if await self._wait_for_voting(candidate_block) is None:
                 return
 
-            if len(candidate_block.body.transactions) == 0 and not conf.ALLOW_MAKE_EMPTY_BLOCK and \
-                    next_leader.hex_hx() != ChannelProperty().peer_id:
+            if self.made_block_count >= (conf.MAX_MADE_BLOCK_COUNT - 1) and next_leader.hex_hx() != ChannelProperty().peer_id:
```
zsaladin: This condition "self.made_block_count >= (conf.MAX_MADE_BLOCK_COUNT - 1)" forces not to change leader even if next_leader is changed. But block has already been made with the "next_leader".

It may not be a problem at the moment, but it is likely to cause problems later.

loopchain/peer/consensus_siever.py
```diff
@@ -98,54 +129,34 @@ def __build_candidate_block(self, block_builder, next_leader, vote_result):
...
+                    """
+                    self._made_block_count += 1
+                elif self.made_block_count >= (conf.MAX_MADE_BLOCK_COUNT - 1):
+                    if last_unconfirmed_block:
+                        await self.__add_block(last_unconfirmed_block)
+                        peer_manager = ObjectManager().channel_service.peer_manager
+                        next_leader = ExternalAddress.fromhex(peer_manager.get_next_leader_peer(
+                            current_leader_peer_id=ChannelProperty().peer_id).peer_id)
+                    else:
+                        raise NotEnoughVotes
```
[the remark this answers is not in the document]
windies21: There is no more right to generate block. Consensus loop have to just return.

loopchain/peer/consensus_siever.py
```diff
@@ -98,54 +129,34 @@ def __build_candidate_block(self, block_builder, next_leader, vote_result):
...
+                elif len(block_builder.transactions) > 0 or conf.ALLOW_MAKE_EMPTY_BLOCK:
+                    if last_unconfirmed_block:
+                        next_leader = await self.__add_block_and_new_epoch(block_builder, last_unconfirmed_block)
+                elif len(block_builder.transactions) == 0 and (
+                        last_unconfirmed_block and len(last_unconfirmed_block.body.transactions) > 0):
+                    next_leader = await self.__add_block_and_new_epoch(block_builder, last_unconfirmed_block)
                 else:
-                    return self.__block_generation_timer.call()
+                    raise NotEnoughVotes
```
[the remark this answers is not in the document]
windies21: You're right. It need same action but It means different situation. So I will change it like this.

....
        raise NothingToMakeBlock
except (NotEnoughVotes, NothingToMakeBlock):
    return self.__block_generation_timer.call()

loopchain/blockchain/blockchain.py
```diff
@@ -661,7 +667,8 @@ def confirm_prev_block(self, current_block: Block):
         #                    f"tx count({len(current_block.body.transactions)}), "
         #                    f"height({current_block.header.height})")
 
-        candidate_blocks = ObjectManager().channel_service.block_manager.candidate_blocks
+        block_manager: BlockManager = ObjectManager().channel_service.block_manager
```
zsaladin: This type hint `BlockManager` at Line 37 is not valid on runtime.
If exceptions is not raised there, another importing `BlockManager` exists. (Maybe import *)

You should block_manager: 'BlockManager' if there is no other importing `BlockManager`.

loopchain/channel/channel_inner_service.py
```diff
@@ -693,11 +693,27 @@ def get_tx_info(self, tx_hash):
...
-        self._channel_service.state_machine.vote(unconfirmed_block=unconfirmed_block)
+        try:
+            self._channel_service.block_manager.verify_confirm_info(unconfirmed_block)
+        except ConfirmInfoInvalid:
+            # TODO
+            pass
+        except ConfirmInfoInvalidNeedBlockSync as e:
+            util.logger.debug(f"ConfirmInfoInvalidNeedBlockSync {e}")
+            block_manager = self._channel_service.block_manager
+            if self._channel_service.state_machine.state == "BlockGenerate" and (
```
zsaladin: Don't leaders block sync?

yakkle:
version update [**commit**](https://github.com/icon-project/loopchain/tree/15e1c3ead26b498a98894684b623ccb9d185e971) is useless.

loopchain/blockchain/blockchain.py
```diff
@@ -678,7 +685,7 @@ def confirm_prev_block(self, current_block: Block):
             except KeyError:
                 if self.last_block.header.hash == current_block.header.prev_hash:
                     logging.warning(f"Already added block hash({current_block.header.prev_hash.hex()})")
-                    if current_block.header.complained:
+                    if current_block.header.complained and block_manager.epoch.complained_result:
```
hangaroy: I think a node can be received valid complained block even if the node have not been received leader complained message.

loopchain/channel/channel_inner_service.py
windies21: Yes. Leader node change its type first and then do block sync next height.

loopchain/blockchain/blockchain.py
windies21: v0.1 block's complained field is not real value. It allow empty block option made same block can treat as a complain block so, It need another checks.

zsaladin (approved):

PR #195, opened 2019-05-20
[LC-363] Take follow-up about removing "blockHash" from receipt

zsaladin:
"blockHash" property will be removed from tx receipt with new Icon Service.
In order to keep compatibility of blockchain DB, "blockHash" property is pushed manually.


zsaladin:
Icon Service has made it compatible by keeping Block Hash in receipt.
So this PR is not needed.

PR #197, opened 2019-05-22
[LC-372] Replace inline signature code with Signer and SignVerifier.

zsaladin:
There are inline signature code in Block and Transaction.
I replaced it with Signer and SignVerifier.


yakkle:
💯

PR #200, opened 2019-05-27
[LC-377] update requirements.txt

metalg0su:
- Prevent python version diverging

[related](https://github.com/icon-project/icon-rpc-server/pull/70)


yakkle:
@metalg0su @hangaroy 
This PR need to update with version up grpc `1.20.1+`

PR #208, opened 2019-06-03
[LC-273] Apply new vote.

zsaladin:
Apply new vote to siever(old)

### DIFF
--- a/loopchain/blockchain/blockchain.py
+++ b/loopchain/blockchain/blockchain.py
@@ -29,6 +29,7 @@ from loopchain.blockchain.blocks import Block, BlockBuilder, BlockSerializer
 from loopchain.blockchain.blocks import BlockProver, BlockProverType, BlockVersioner
 from loopchain.blockchain.transactions import Transaction, TransactionBuilder
 from loopchain.blockchain.transactions import TransactionSerializer, TransactionVersioner
+from loopchain.blockchain.votes.v0_1a import BlockVotes, LeaderVotes
 from loopchain.blockchain.exception import *
 from loopchain.blockchain.score_base import *
 from loopchain.channel.channel_property import ChannelProperty
@@ -342,7 +343,7 @@ class BlockChain:
         if confirm_info:
             batch.Put(
                 BlockChain.CONFIRM_INFO_KEY + block_hash_encoded,
-                b'0x1'
+                json.dumps(confirm_info.serialize()).encode("utf-8")
             )
 
         self.__confirmed_block_db.Write(batch)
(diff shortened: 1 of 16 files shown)


loopchain/blockchain/blockchain.py
```diff
@@ -705,7 +708,8 @@ def confirm_prev_block(self, current_block: Block):
 
             # utils.logger.debug(f"-------------------confirm_prev_block---before add block,"
             #                    f"height({unconfirmed_block.header.height})")
-            self.add_block(unconfirmed_block, current_block.body.confirm_prev_block)
+            confirm_info = current_block.body.prev_votes if current_block.header.version == "0.3" else None
```
hyerim-kim: Is the value of `None` fine for other versions to except for v0.3?
I wonder why you don't use `current_block.body.confirm_prev_block` for other versions.

loopchain/blockchain/blocks/block_verifier.py
```diff
@@ -99,18 +99,18 @@ def verify_version(self, block: 'Block'):
...
                                      f"Height({block.header.height}), "
                                      f"Expected({prev_block.header.height + 1}).")
             self._handle_exception(exception)
 
+        if block.header.prev_hash != prev_block.header.hash:
+            exception = RuntimeError(f"Block({block.header.height}, {block.header.hash.hex()}, "
+                                     f"PrevHash({block.header.prev_hash.hex()}), "
+                                     f"Expected({prev_block.header.hash.hex()}).")
+            self._handle_exception(exception)
+
```
hyerim-kim: Let me know what the difference of the order of conditional code means.

zsaladin: It seems an unnecessary change.
I will remove it.

loopchain/peer/consensus_siever.py
```diff
@@ -94,7 +96,7 @@ def __build_candidate_block(self, block_builder, next_leader, vote_result):
         if not vote_result:
             raise NotEnoughVotes
 
-        self._block_manager.get_blockchain().add_block(block, vote)
+        self._block_manager.get_blockchain().add_block(block, vote.votes)
```
yakkle: @windies21 @hyerim-kim @zsaladin 

in `blockchain.py`,
```
    def add_block(self,
                  block: Block,
                  confirm_info=None,
                  need_to_write_tx_info=True,
                  need_to_score_invoke=True) -> bool:
```
`add_block` has confirm_info as second parameter.

if `vote.votes` is confirm_info, it would be better to add comment above.

PR #212, opened 2019-06-04
 [LC-397] change return type of channel info detail

laurenceyoon:
This pull request should be merged after LC-386 is merged to develop.


zsaladin (approved):

loopchain/channel/channel_service.py
```diff
@@ -194,18 +194,17 @@ def cleanup(self):
             self.__timer_service.wait()
             logging.info("Cleanup TimerService.")
 
-    async def init(self, peer_port, peer_target, rest_target, radio_station_target, peer_id,
-                   group_id_will_removed, node_type, score_package):
-        loggers.get_preset().peer_id = peer_id
+    async def init(self, **kwargs):
```
yakkle: It would be better to add docs of `kwargs` parameter that description of one's keys.
Or change to key-value parameters.

yakkle (approved):

PR #225, opened 2019-06-14
[LC-412] Issues caused by merging Block(0.3) and Prep list

zsaladin:
- Fix not being able to find prev vote.
- switch the order of init_network and ready_to_height_sync
- Apply removing blockHash to block prover on receipt

### DIFF
--- a/loopchain/channel/channel_service.py
+++ b/loopchain/channel/channel_service.py
@@ -195,6 +195,12 @@ class ChannelService:
             logging.info("Cleanup TimerService.")
 
     async def init(self, **kwargs):
+        """Initialize Channel Service
+
+        :param kwargs: takes (peer_id, peer_port, peer_target, rest_target, rs_target, node_type, score_package)
+        within parameters
+        :return: None
+        """
         loggers.get_preset().peer_id = kwargs.get('peer_id')
         loggers.get_preset().update_logger()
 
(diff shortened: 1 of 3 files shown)


loopchain/channel/channel_service.py
```diff
@@ -652,7 +649,9 @@ def __ready_to_height_sync(self):
         blockchain = self.block_manager.get_blockchain()
 
         blockchain.init_blockchain()
-        if blockchain.block_height == -1 and 'genesis_data_path' in conf.CHANNEL_OPTION[ChannelProperty().name]:
+        if blockchain.block_height == -1 \
+                and 'genesis_data_path' in conf.CHANNEL_OPTION[ChannelProperty().name] \
+                and self.is_support_node_function(conf.NodeFunction.Vote):
```
yakkle: if condition statement is too long, how about create check method that return `bool`.
for example,
```py
def _is_genesis(self):
    if condition1 and condtion2 and condtion3:
        return true
    return false
...
in another method
...

    if self._is_genesis():
        generate_genesis_block()
```

yakkle: in addition to, I suggest use parenthesis with multiline conditions.
```py
if (condition1
       and condition2
       and condition3):
    # do something..
```

PR #228, opened 2019-06-18
[LC-414] fix bug on BlockHeightSync and Watch state and add timeout on jsonrpc call

laurenceyoon:
- close websocket connection when AnnounceNewBlockError raises.
- remove repeat timer in block_sync_timer (30s)
- add timeout(30s) on jsonrpc call. 
- reject already registered citizen when parent node got new register request.

### DIFF
--- a/loopchain/baseservice/node_subscriber.py
+++ b/loopchain/baseservice/node_subscriber.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Class for websocket client between child node and RS peer"""
+"""Class for websocket client between child node and parent node"""
 
 import asyncio
 import json
@@ -27,7 +27,7 @@ from jsonrpcserver.aio import AsyncMethods
 from loopchain import configure as conf
 from loopchain import utils
 from loopchain.baseservice import ObjectManager, TimerService, Timer
-from loopchain.blockchain import BlockSerializer, BlockVerifier
+from loopchain.blockchain import BlockSerializer, BlockVerifier, AnnounceNewBlockError
 from loopchain.channel.channel_property import ChannelProperty
 from loopchain.protos import message_code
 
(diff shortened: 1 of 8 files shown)


loopchain/baseservice/node_subscriber.py
```diff
@@ -40,49 +40,50 @@
...
             # set websocket payload maxsize to 4MB.
-            self.__websocket = await websockets.connect(self.__target_uri, max_size=4 * conf.MAX_TX_SIZE_IN_BLOCK)
+            self._websocket = await websockets.connect(self._target_uri, max_size=4 * conf.MAX_TX_SIZE_IN_BLOCK)
             logging.debug(f"Websocket connection is Completed.")
             request = Request("node_ws_Subscribe", height=block_height, peer_id=ChannelProperty().peer_id)
-            await self.__websocket.send(json.dumps(request))
-            await self.__subscribe_loop(self.__websocket)
+            await self._websocket.send(json.dumps(request))
+            await self._subscribe_loop(self._websocket)
+        except AnnounceNewBlockError as e:
```
zsaladin: You don't have to use 'e'
```python
except AnnounceNewBlockError:
    raise
```

loopchain/peer/block_manager.py
```diff
@@ -298,7 +298,7 @@ def add_confirmed_block(self, confirmed_block: Block, confirm_info=None):
             if result:
                 return
 
-        self.block_height_sync()
+        self.__channel_service.state_machine.block_sync()
```
zsaladin: How about raising an exception instead of block sync.
Determining if it needs to sync blocks is placed out of this method.

loopchain/peer/block_manager.py
```diff
@@ -370,19 +370,15 @@ def __block_request_by_citizen(self, block_height, rs_rest_stub):
                 'height': str(block_height)
             }
         )
-        max_height_result = rs_rest_stub.call("Status")
-
-        if max_height_result.status_code != 200:
-            raise ConnectionError
-
+        last_height = rs_rest_stub.call("GetLastBlock").get('height')
```
zsaladin: Deserialization is needed.
Row 'height' can be hex string.

loopchain/baseservice/node_subscriber.py
laurenceyoon: Resolved.

loopchain/peer/block_manager.py
laurenceyoon: Resolved.

loopchain/peer/block_manager.py
laurenceyoon: This line is removed, expecting channel_service would handle the exception during publish_new_block in node_subscriber.

loopchain/baseservice/rest_stub_manager.py
```diff
@@ -13,13 +13,12 @@
 # limitations under the License.
 """A stub wrapper for REST call.
 This object has same interface with gRPC stub manager"""
-
+import json
```
yakkle: Is this using?

laurenceyoon: Nope! removed:)

PR #234, opened 2019-06-24
[LC-429] Fix handling exceptions on block sync

zsaladin:
- Fix handling exceptions on block sync

### DIFF
--- a/loopchain/peer/block_manager.py
+++ b/loopchain/peer/block_manager.py
@@ -485,20 +485,23 @@ class BlockManager:
                                       self.blockchain.last_block,
                                       self.blockchain,
                                       reps_getter=reps_getter)
-        need_to_write_tx_info, need_to_score_invoke = True, True
+
+        tx_duplicated_exceptions = []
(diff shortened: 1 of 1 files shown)


loopchain/peer/block_manager.py
```diff
@@ -514,21 +514,25 @@ def __add_block_by_sync(self, block_, confirm_info=None):
...
-                raise exc
+                other_exceptions.append(exc)
+
+        if other_exceptions:
+            raise other_exceptions[0]
+        if score_invoke_exceptions and not tx_duplicated_exceptions:
+            raise score_invoke_exceptions[0]
 
+        need_to_write_tx_info = not tx_duplicated_exceptions
+        need_to_score_invoke = not score_invoke_exceptions
```
laurenceyoon: I think this line should be like this (to maintain the same action as before):

```py
need_to_score_invoke = not score_invoke_exceptions or need_to_write_tx_info
```

metalg0su (approved):

yakkle:
stale

PR #245, opened 2019-07-04
[LC-452] Load `confirm_info` from DB when a node restarts.

hyerim-kim:
- Load `confirm_info` from DB when a node restarts.
- Edit the order of logic to verify `height` earlier than `prev_hash` in `verify_prev_block`.

### DIFF
--- a/loopchain/blockchain/blocks/block_versioner.py
+++ b/loopchain/blockchain/blocks/block_versioner.py
@@ -2,6 +2,7 @@ import json
 from collections import namedtuple
 from typing import List, Union
 
+from loopchain.blockchain.types import Hash32
 BlockVersion = namedtuple("BlockVersion", ("height", "name"))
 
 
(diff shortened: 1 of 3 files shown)


loopchain/blockchain/blocks/block_versioner.py
```diff
@@ -37,6 +38,13 @@ def get_height(self, block_dumped: Union[str, dict]):
         height = block_dumped["height"]
         return int(height, 16) if isinstance(height, str) else height
 
+    def get_hash(self, block_dumped: Union[str, dict]):
+        if isinstance(block_dumped, str):
+            block_dumped = json.loads(block_dumped)
+
+        _hash = block_dumped["block_hash"]
```
zsaladin: 1. It is not compatible with block version 0.3. It has "hash"
2. Appending single underscore is better to avoid conflicting reversed words. Or just use 'block_hash'

zsaladin (approved):

windies21 (approved):
LGTM

PR #255, opened 2019-07-16
[LC-462] support multichannel of node_type

laurenceyoon:
- set default service_type as 'citizen'
- manage node_type in channel_service, not in peer_service
- set CommunityNode if rs_target is None.

### DIFF
--- a/loopchain/channel/channel_service.py
+++ b/loopchain/channel/channel_service.py
@@ -184,7 +184,7 @@ class ChannelService:
     async def init(self, **kwargs):
         """Initialize Channel Service
 
-        :param kwargs: takes (peer_id, peer_port, peer_target, rest_target, rs_target, node_type)
+        :param kwargs: takes (peer_id, peer_port, peer_target, rest_target, rs_target)
         within parameters
         :return: None
         """
@@ -196,7 +196,12 @@ class ChannelService:
         ChannelProperty().rest_target = kwargs.get('rest_target')
         ChannelProperty().radio_station_target = kwargs.get('rs_target')
         ChannelProperty().peer_id = kwargs.get('peer_id')
-        ChannelProperty().node_type = conf.NodeType(kwargs.get('node_type'))
+
+        # FIXME this is temporary setting for node_type.
+        if ChannelProperty().radio_station_target:
+            ChannelProperty().node_type = conf.NodeType.CitizenNode
+        else:
+            ChannelProperty().node_type = conf.NodeType.CommunityNode
 
         self.__peer_manager = PeerManager()
         await self.__init_peer_auth()
@@ -208,7 +213,8 @@ class ChannelService:
         await self.__init_sub_services()
 
     async def __init_network(self):
-        self.__init_radio_station_stub()
+        if ChannelProperty().radio_station_target:
+            self.__init_radio_station_stub()
         await self._load_peers()
 
     async def evaluate_network(self):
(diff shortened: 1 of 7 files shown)


loopchain/utils/command_arguments/__init__.py
```diff
@@ -65,7 +67,7 @@ def command(self, value):
 
 attributes = {
     Type.ServiceType:
-        Attribute("service_type", type=str,
+        Attribute("service_type", type=str, default='citizen', nargs=OPTIONAL,
```
yakkle: Why don't you use `'?'` instead of `OPTIONAL` constant?
No more usage `OPTIONAL` in this scope and if you need more `nargs` in future, it would be better to provide various options(`?, +, *, etc...`)

loopchain/launcher.py
```diff
@@ -230,12 +228,12 @@ def start_as_tool(args):
...
 
 
-def start_as_peer(args, node_type=None):
+def start_as_peer(args):
     print_prologue()
 
     # apply default configure values
     port = args.port or conf.PORT_PEER
-    radio_station_target = f"{conf.IP_RADIOSTATION}:{conf.PORT_RADIOSTATION}"
+    radio_station_target = None
```
yakkle: if `radio_station_target` is None, no more needed.

loopchain/peer/peer_service.py
```diff
@@ -348,16 +308,3 @@ def close(self):
 
     def _load_channel_infos(self):
         self._channel_infos = self._get_channel_infos()
-
-    async def change_node_type(self, node_type):
```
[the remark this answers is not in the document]
laurenceyoon: The PeerService no longer manages the node_type, so it doesn't matter whether the node_type has changed or not. When node_type is changed, it is managed by the corresponding channel because it is related only to a specific channel.

loopchain/utils/command_arguments/__init__.py
laurenceyoon: OK. I thought it might be easier to recognize if it's set with specific meaning.

loopchain/launcher.py
laurenceyoon: But the citizen nodes still needs radio_station_target.

yakkle: aha! you are right.

zsaladin (approved):

loopchain/utils/command_arguments/__init__.py
```diff
@@ -65,7 +65,7 @@ def command(self, value):
 
 attributes = {
     Type.ServiceType:
-        Attribute("service_type", type=str,
+        Attribute("service_type", type=str, default='citizen', nargs='?',
                   help="loopchain service to start [rs|peer|citizen|rest|rest-rs|score|tool|admin]"),
```
metalg0su: Are we use all of NodeTypes written in help msg?
If not, it is better to rewrite message.

loopchain/channel/channel_service.py
```diff
@@ -348,7 +353,7 @@ def __init_broadcast_scheduler(self):
         scheduler.schedule_job(BroadcastCommand.SUBSCRIBE, ChannelProperty().peer_target,
                                block=True, block_timeout=conf.TIMEOUT_FOR_FUTURE)
 
-    def __init_radio_station_stub(self):
+    async def __init_radio_station_stub(self):
```
metalg0su: async keyword is needed?

loopchain/utils/command_arguments/__init__.py
laurenceyoon: You're right. `rs, rest rest-rs` won't be needed anymore.

loopchain/channel/channel_service.py
laurenceyoon: Checked. I'll fix this.

PR #267, opened 2019-07-31
[LC-495] add api get reps by rep root hash.

windies21:


loopchain/blockchain/blockchain.py
```diff
@@ -71,19 +74,19 @@ def __init__(self, channel_name=None, store_id=None):
...
+        self.__block_manager: BlockManager = block_manager
 
         store_id = f"{store_id}_{channel_name}"
-        # block db has [ block_hash - block | block_height - block_hash | BlockChain.LAST_BLOCK_KEY - block_hash ]
         self._blockchain_store, self._blockchain_store_path = utils.init_default_key_value_store(store_id)
 
-        # made block count as a leader
-        self.__invoke_results = {}
+        # tx receipts and next prep after invoke, {block_hash: (receipts, next_prep)}
+        self.__invoke_results: AgingCache = AgingCache(max_age_seconds=conf.INVOKE_RESULT_AGING_SECONDS)
```
zsaladin: 1. How about using Hash32 for BlockHash not Hash32.hex()?

2. Handling it with time seems uncertain.
I recommend
```python
self._tx_results_cache: DefaultDict[int, Dict[Hash32, (dict, Sequence[ExternalAddress])]] = defaultdict(dict)  
receipts, next_preps = self._receipt_cache[block_height][block_hash]  # use case

def trim_cache(self, block_height: int):
    remove_heights = [height_cache for height_cache in self._tx_results_cache
                                     if height_cache <= block_height]
    for remove_height in remove_heights:
        self._tx_results_cache.pop(remove_height, pop)

File: loopchain/blockchain/blockchain.py [hunk 2]
```diff
@@ -71,19 +74,19 @@ def __init__(self, channel_name=None, store_id=None):
...
 
         self.__add_block_lock = threading.RLock()
         self.__confirmed_block_lock = threading.RLock()
 
         self.__total_tx = 0
-        self.__nid: str = None
+        self.__nid: Optional[str] = None
```
zsaladin: Should we use Optional since now when variable can be None?

loopchain/blockchain/blockchain.py
```diff
@@ -269,25 +272,21 @@ def add_block(self,
...
         with self.__add_block_lock:
-            invoke_results = self.__invoke_results.get(block.header.hash.hex(), None)
-            if invoke_results is None and need_to_score_invoke:
-                if block.header.height == 0:
-                    block, invoke_results = ObjectManager().channel_service.genesis_invoke(block)
-                else:
-                    block, invoke_results = ObjectManager().channel_service.score_invoke(block, self.__last_block)
+            receipts, next_prep = self.__invoke_results.get(block.header.hash.hex(), (None, None))
+            if receipts is None and need_to_score_invoke:
+                self.get_invoke_func(block.header.height)(block, self.__last_block)
```
zsaladin: Are you sure block.prev_hash ==  self.__last_block.hash?
What if the condition won't be meet?

loopchain/blockchain/blockchain.py
```diff
@@ -269,25 +272,21 @@ def add_block(self,
...
 
             try:
-                if need_to_write_tx_info:
-                    self.__add_tx_to_block_db(block, invoke_results)
+                if not need_to_write_tx_info:
```
zsaladin: Is there any case that we don't need to write tx info anymore?

It was because of writing tx results before 'score write commit'.
Above logic is removed now. So this flag has become useless.

loopchain/blockchain/blockchain.py
```diff
@@ -269,25 +272,21 @@ def add_block(self,
...
+                    receipts = None
                 if need_to_score_invoke:
```
zsaladin: Is there any case that we don't need to score invoke?
Same reason I wrote above.

loopchain/blockchain/blockchain.py
```diff
@@ -269,25 +272,21 @@ def add_block(self,
...
                     ObjectManager().channel_service.score_write_precommit_state(block)
+                next_total_tx = self.__write_block_data(block, confirm_info, receipts, next_prep)
```
zsaladin: Write block data has to be place before 'write precommit state' so that we can recover it easily on crashing of 'write precommit state'

loopchain/blockchain/blockchain.py
```diff
@@ -317,7 +316,51 @@ def __add_block(self, block: Block, confirm_info, need_to_write_tx_info=True, ne
 
             return True
 
-    def __write_block_data(self, block: Block, confirm_info):
+    def __add_tx_to_block_db(self, block, receipts, batch=None):
+        """block db 에 block_hash - block_object 를 저장할때, tx_hash - block_hash 를 저장한다.
```
zsaladin: Writing it in english is needed.

loopchain/blockchain/blockchain.py
```diff
@@ -317,7 +316,51 @@ def __add_block(self, block: Block, confirm_info, need_to_write_tx_info=True, ne
...
+            }
+
+            write_target.put(
+                tx_hash.encode(encoding=conf.HASH_KEY_ENCODING),
+                json.dumps(tx_info).encode(encoding=conf.PEER_DATA_ENCODING))
+
+            tx_queue.pop(tx_hash, None)
+
+            if block.header.height > 0:
+                self.__save_tx_by_address(tx)
```
zsaladin: Support it? V2 has been deprecated.

loopchain/blockchain/blockchain.py
```diff
@@ -317,7 +316,51 @@ def __add_block(self, block: Block, confirm_info, need_to_write_tx_info=True, ne
 
             return True
 
-    def __write_block_data(self, block: Block, confirm_info):
+    def __add_tx_to_block_db(self, block, receipts, batch=None):
```
zsaladin: Changinge name is needed. It should be similar to 'write_block_data'.
Maybe 'write_transtions'? Remove 'block_db'

loopchain/blockchain/blockchain.py
windies21: Hash32 is a good suggestion. And this is not a cache. In normal operation, invoke_result always should be empty when processed. But I will consider your recommend later.

loopchain/blockchain/blockchain.py
windies21: I'm not sure. I just hated the IDE warning. How's your  opinion?

loopchain/blockchain/blockchain.py
windies21: I think, It's a same logic as before. There may be a problem but this is not a concern in this PR.

loopchain/blockchain/blockchain.py
windies21: It is a kind of trick to keep the program logic the same as before. It is not a modification of the flow here.

windies21: 예전 코드는 need_to_write_tx_info 가 있을때에만 add_tx_to_block_db 를 수행하도록 되어 있었기에 같은 흐름을 만들기 위해서 write_block_data 에 receipts 를 None 으로 전달하게 하여 동일하게 동작하게 한 것입니다. 이 곳의 코드 로직을 개선한 것은 아닙니다. 이 PR은 논리적인 리팩토링을 한것이 아니라 전적으로 같은 흐름을 유지하면서 invoke 함수들의 위치만 channel_service 에서 blockchain 으로 옮겨왔습니다. 이유는 invoke 할때 생성되는 정보를 blockchain 이 __invoke_result 로 관리를 하고 있는데 여기에 invoke 후 복수의 값을 저장할 필요가 생겼기 때문입니다.

loopchain/blockchain/blockchain.py
```diff
@@ -340,6 +383,17 @@ def __write_block_data(self, block: Block, confirm_info):
...
+            batch = self.__add_tx_to_block_db(block, receipts, batch)
+
+        if next_prep:
+            utils.logger.notice(f"store next_prep in __write_block_data\nprep_hash({next_prep['rootHash']})"
+                                f"\npreps({next_prep['preps']})")
+            batch.put(
+                BlockChain.PREPS_KEY + next_prep['rootHash'].encode(encoding=conf.HASH_KEY_ENCODING),
+                json.dumps(next_prep['preps']).encode(encoding=conf.PEER_DATA_ENCODING)
+            )
+
```
laurenceyoon: Is there any information about `startTermHeight` and `endTermHeight`?

loopchain/blockchain/blockchain.py
windies21: the program logic the same as before. not  concern in this PR.

loopchain/blockchain/blockchain.py
windies21: 이 PR에서 고민한 내용은 아닙니다. 기존 코드 동작으로 block mismatch 가 구현되어 있고 테스트되었을텐데 이 코드의 순서의 논리적인 문제점을 고민한 PR은 아닙니다. 그냥 기존과 동일한 순서상에 위치시켰습니다. 하지만 이 문제를 나중에 고민해볼 필요는 있을 것 같습니다.

loopchain/blockchain/blockchain.py
windies21: I just keep it as a same before. I will consider it later refactoring.

loopchain/blockchain/blockchain.py
windies21: I didn't write it in this PR. but I will change it.

loopchain/blockchain/blockchain.py
windies21: It's not possible support now.

loopchain/blockchain/blockchain.py
windies21: However, no matter how I think, the order seemed strange, so I fixed it. review again please.

loopchain/blockchain/blockchain.py
zsaladin: Assume that a situation invoking many times with diffent block hash in one block height.

Then only 'invoke result' of added block will be removed. Other invoke results remains until the node shuts down.

loopchain/blockchain/blockchain.py
zsaladin: I prefer to use `Optional` in this case.

loopchain/blockchain/blockchain.py
zsaladin: But as you mentioned this PR is not considered of refactoring. Then just keep it . We will be able to improve it later.

zsaladin:
Codes in this PR should be refactored. We will keep some issues mentioned above.

loopchain/blockchain/blockchain.py
windies21: OK. I'm agree with you.

PR #269, opened 2019-08-01
[LC-501] parse unknown argument as a quick command if it's possible.

windies21:
* parse unknown argument as a quick command if it's possible.


yakkle:
retest this please

PR #270, opened 2019-08-02
[LC-503] Save peer manger's peer list as a preps data by its hash, before revision-6 for compatibility. 

windies21:


loopchain/blockchain/epoch.py
```diff
@@ -115,35 +115,6 @@ def complain_result(self) -> Optional[str]:
         else:
             return None
 
-    def pop_complained_candidate_leader(self):
```
windies21: What is the purpose of this function? Anyone know?

loopchain/blockchain/blockchain.py
```diff
@@ -237,6 +237,21 @@ def find_confirm_info_by_height(self, height) -> bytes:
...
         return bytes()
 
+    def find_preps_by_roothash(self, roothash: str) -> bytes:
+        hash_encoded = roothash.encode(encoding='UTF-8')
+        try:
+            return bytes(self._blockchain_store.get(BlockChain.PREPS_KEY + hash_encoded))
+        except KeyError:
+            return bytes()
+
+    def save_preps(self, roothash: str, preps: dict, batch=None):
```
metalg0su: I think `batch` argument could raise misunderstanding.
How about add type hint `KeyValueStoreWriteBatch` or something..?

```suggestion
    def save_preps(self, roothash: str, preps: dict, batch: KeyValueStoreWriteBatch = None):
```

loopchain/blockchain/blockchain.py
```diff
@@ -237,6 +237,21 @@ def find_confirm_info_by_height(self, height) -> bytes:
 
         return bytes()
 
+    def find_preps_by_roothash(self, roothash: str) -> bytes:
```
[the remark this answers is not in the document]
windies21: Your suggestion is very reasonable, but the name comes from the IISS. So I think it is less confusing to keep that name here.

loopchain/baseservice/peer_manager.py
```diff
@@ -37,34 +37,28 @@
...
+    """Manage peer list as serializable data."""
     def __init__(self):
-        self.peer_list: Dict[BlockHeader.peer_id, Peer] = {}  # { peer_id:Peer }
-        self.peer_order_list: Dict[int, BlockHeader.peer_id] = {}  # { order:peer_id }
-        self.peer_leader: int = 0  # leader_order
+        self.peer_list: OrderedDict[BlockHeader.peer_id, Peer] = OrderedDict()  # { peer_id:Peer }
+        self.leader_id: BlockHeader.peer_id = None  # leader_peer_id
 
     def serialize(self) -> dict:
         peer_list_serialized = {peer_id: peer.serialize()
```
zsaladin: This can't keep the order of peer.

loopchain/baseservice/peer_manager.py
```diff
@@ -37,34 +37,28 @@
...
     @staticmethod
     def deserialize(peer_list_data_serialized: dict) -> 'PeerListData':
         peer_list = {peer_id: Peer.deserialize(peer_serialized)
                      for peer_id, peer_serialized in peer_list_data_serialized['peer_list'].items()}
 
-        peer_order_list = {int(order): peer_id
-                           for order, peer_id in peer_list_data_serialized['peer_order_list'].items()}
-
         peer_list_data = PeerListData()
         peer_list_data.peer_list = peer_list
```
zsaladin: Ordering can't kept

loopchain/baseservice/peer_manager.py
```diff
@@ -76,59 +70,62 @@ def load(peer_list_data_dumped: bytes):
...
-        return self.get_peer_ids_hash(peer_ids)
-
-    @staticmethod
-    def get_peer_ids_hash(peer_ids):
-        peer_ids_hash = hashlib.sha256(peer_ids.encode(encoding='UTF-8')).hexdigest()
-        util.logger.debug(f"peer ids hash({peer_ids_hash})")
-        return peer_ids_hash
+    def serialize_as_preps(self) -> list:
+        return [{'id': peer_id, 'p2pEndpoint': peer.target}
+                for peer_id, peer in self._peer_list_data.peer_list.items()]
```
laurenceyoon: Since `p2pEndpoint` name is from IISS API, shouldn't we use `address` for prep's id as well?

laurenceyoon: `id` would be fine, because `address` is for external API.

PR #288, opened 2019-08-21
[LC-516] Prevent the occurrence of `TransactionDuplicatedError` and process the case that voters of quorum or above vote against a block.

hyerim-kim:
Currently, a leader can bond invalid block and occur a problem even though voters of quorum or above vote against a block. That's why I add logic to process by a leader to avoid bonding any invalid block in Siever. And I add logic to prevent the occurrence of `TransactionDuplicatedError` because `tx_info` doesn't be stored if the block, it includes the transaction of `tx_info`, has been never invoked.

### DIFF
--- a/loopchain/blockchain/exception.py
+++ b/loopchain/blockchain/exception.py
@@ -14,15 +14,15 @@
 """A module of exceptions for errors on block chain"""
 
 from typing import TYPE_CHECKING
-from loopchain.protos import message_code
 
+from loopchain.protos import message_code
 
 if TYPE_CHECKING:
     from loopchain.blockchain.transactions import Transaction
     from loopchain.blockchain.types import Hash32
 
 
-class BlockInValidError(Exception):
+class InvalidBlock(Exception):
     """블럭 검증오류
     검증되지 않은 블럭이 블럭체인에 추가되거나, 검증시 hash값이 다르다면 발생한다
     """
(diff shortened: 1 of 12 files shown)


loopchain/blockchain/exception.py
```diff
@@ -14,15 +14,15 @@
...
+from loopchain.protos import message_code
 
 if TYPE_CHECKING:
     from loopchain.blockchain.transactions import Transaction
     from loopchain.blockchain.types import Hash32
 
 
-class BlockInValidError(Exception):
+class InvalidBlock(Exception):
     """블럭 검증오류
```
windies21: how about comment in english?

loopchain/peer/consensus_siever.py
```diff
@@ -96,36 +96,38 @@ def __build_candidate_block(self, block_builder, next_leader):
...
         self._block_manager.candidate_blocks.remove_block(block.header.hash)
         self._blockchain.last_unconfirmed_block = None
 
+    def __get_complaint_votes(self):
+        if self._block_manager.epoch.round > 0:
+            return self._block_manager.epoch.complain_votes[self._block_manager.epoch.round - 1]
+        return None
+
     async def consensus(self):
+        util.logger.debug(f"-------------------consensus-------------------")
```
windies21: develop branch has this comment already. see develop when rebase with it.

zsaladin (approved):

loopchain/peer/consensus_siever.py
hyerim-kim: 👌

windies21 (approved):

zsaladin (approved):

PR #299, opened 2019-08-28
[LC-551] set prevBlockValidators value with ids from votes

windies21:
* soobok Jin forever~~~

### DIFF
--- a/loopchain/blockchain/blockchain.py
+++ b/loopchain/blockchain/blockchain.py
@@ -1082,6 +1082,15 @@ class BlockChain:
             }
             transactions.append(transaction)
 
+        if prev_block.header.version != "0.1a":
+            prev_block_validators = [vote.rep.hex_hx() for vote in _block.body.prev_votes
+                                     if vote.rep != prev_block.header.peer_id]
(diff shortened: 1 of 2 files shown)


loopchain/blockchain/blockchain.py
```diff
@@ -1082,6 +1082,18 @@ def score_invoke(self, _block: Block, prev_block: Block, is_block_editable: bool
             }
             transactions.append(transaction)
 
+        if prev_block.header.version != "0.1a":
+            prev_block_validators = \
+                [vote.rep.hex_hx()
+                 for vote in _block.body.prev_votes
+                 if vote.rep != prev_block.header.peer_id]
```
zsaladin: Why do you exclude leader from reps?

soobokjin: The reason is for the IISS reward. In the icon service, we distribute reward (i.e. I-SCORE) according to the role of P-Rep. For example, The P-Rep who generated the block gets more reward than the P-Reps who validated the block. To distinguish this, the generator P-Rep is excluded from the validator P-Reps.

PR #310, opened 2019-09-04
[LC-568] WIP: Channel service graceful shutdown

metalg0su:


loopchain/channel/channel_service.py
```diff
@@ -135,12 +135,10 @@ def serve(self):
         loop = MessageQueueService.loop
         # loop.set_debug(True)
         loop.create_task(_serve())
-        loop.add_signal_handler(signal.SIGINT, self.close)
-        loop.add_signal_handler(signal.SIGTERM, self.close)
```
yakkle: also, cleanup before `loop.close()` I think.

PR #311, opened 2019-09-04
[LC-571] Fix bug when citizen node become a leader at the prep change.

windies21:


loopchain/peer/block_manager.py
```diff
@@ -628,15 +628,7 @@ def _handle_exception(e):
...
-        current_block_header = self.__current_last_block().header
-        current_height = current_block_header.height
-        next_leader = current_block_header.next_leader
-        leader_peer = self.__channel_service.peer_manager.get_peer(next_leader.hex_hx()) if next_leader else None
-
-        if leader_peer:
-            self.epoch = Epoch.new_epoch(leader_peer.peer_id)
-        elif self.epoch and self.epoch.height < current_height:
-            self.epoch = Epoch.new_epoch()
+        self.epoch = Epoch.new_epoch()
```
hyerim-kim: I think there is nothing different between `start_epoch` and `self.epoch = Epoch.new_epoch()`.

windies21: I understood your point. This part will be changed in the next PR.

hyerim-kim: 👌

PR #314, opened 2019-09-07
[LC-582] Set block header attributes according to P-Rep list status.

metalg0su:
# Goal
Set block header attributes according to P-Rep list status.

# Detail
Suppose that 3 cases when build & invoke block.
Each of cases has different block header attributes. Explain below.

## In Case Normal
No changes in P-Rep list

- next_leader: No changes (TODO: ignored when leader-complained?)
- reps_hash: No changes
- next_reps_hash: Set empty hash

## In Case Term End (Reset order)
P-Rep term expired and switched to new list

- next_leader: Set empty hash
- reps_hash: No changes
- next_reps_hash: Invoked value of `next prep root hash` from score

## In case Panelty (Keep Order)
One of P-Rep has been punished and excluded from list 

- next_leader: No changes
- reps_hash: No changes
- next_reps_hash: Invoked value of `next prep root hash` from score

### DIFF
--- a/loopchain/baseservice/score_code.py
+++ b/loopchain/baseservice/score_code.py
@@ -28,4 +28,4 @@ class ScoreResponse(IntEnum):
 
 class PrepChangedReason:
     TERM_END = "0x0"  # Term ends and reset P-Rep list
-    PANELTY = "0x1"  # Updated within term (Panelty)
+    PENALTY = "0x1"  # Updated within term (Penalty)
--- a/loopchain/baseservice/score_code.py
+++ b/loopchain/baseservice/score_code.py
@@ -24,3 +24,8 @@ class ScoreResponse(IntEnum):
     NOT_INVOKED = 2  # means pending
     NOT_EXIST = 3  # considered fail
     SCORE_CONTAINER_EXCEPTION = 9100
+
+
+class PrepChangedReason:
+    TERM_END = "0x0"  # Term ends and reset P-Rep list
+    PANELTY = "0x1"  # Updated within term (Panelty)
--- a/loopchain/blockchain/blockchain.py
+++ b/loopchain/blockchain/blockchain.py
@@ -123,12 +123,6 @@ class BlockChain:
     def just_before_max_made_block_count(self) -> bool:
         return self.leader_made_block_count == (conf.MAX_MADE_BLOCK_COUNT - 1)
 
-    @property
-    def last_block_has_changed_next_reps(self) -> bool:
-        if self.__last_block.header.version != '0.1a':
-            return self.__last_block.header.prep_changed is not None
-        return False
-
     def _increase_made_block_count(self, block: Block) -> None:
         """This is must called before changing self.__last_block!
 
@@ -138,7 +132,7 @@ class BlockChain:
         if block.header.height == 0:
             return
 
-        if self.__last_block.header.peer_id != block.header.peer_id or self.last_block_has_changed_next_reps:
+        if self.__last_block.header.peer_id != block.header.peer_id or self.__last_block.header.prep_changed:
             self.__made_block_counter[block.header.peer_id] = 1
         else:
             self.__made_block_counter[block.header.peer_id] += 1
@@ -839,12 +833,6 @@ class BlockChain:
 
         return results
 
-    def __is_1st_block_of_new_term(self, unconfirmed_block_header):
-        if unconfirmed_block_header.version == '0.1a':
-            return False
-
-        return unconfirmed_block_header.prep_changed
-
     def confirm_prev_block(self, current_block: Block):
         """confirm prev unconfirmed block by votes in current block
 
@@ -869,7 +857,7 @@ class BlockChain:
                 if self.last_block.header.hash == current_block.header.prev_hash:
                     logging.warning(f"Already added block hash({current_block.header.prev_hash.hex()})")
                     if (current_block.header.complained and self.__block_manager.epoch.complained_result)\
-                            or self.__is_1st_block_of_new_term(self.last_block.header):
+                            or self.last_block.header.prep_changed is not None:
                         utils.logger.debug("reset last_unconfirmed_block by complain block or first block of new term.")
                         self.last_unconfirmed_block = current_block
                     return None
--- a/loopchain/blockchain/blockchain.py
+++ b/loopchain/blockchain/blockchain.py
@@ -126,7 +126,7 @@ class BlockChain:
     @property
     def last_block_has_changed_next_reps(self) -> bool:
         if self.__last_block.header.version != '0.1a':
-            return self.__last_block.header.reps_hash != self.__last_block.header.next_reps_hash
+            return self.__last_block.header.prep_changed is not None
         return False
 
     def _increase_made_block_count(self, block: Block) -> None:
@@ -149,8 +149,8 @@ class BlockChain:
     def get_first_leader_of_next_reps(self) -> str:
         utils.logger.notice(
             f"in get_next_leader new reps leader is "
-            f"{self.find_preps_ids_by_roothash(self.last_block.header.next_reps_hash)[0]}")
-        return self.find_preps_ids_by_roothash(self.last_block.header.next_reps_hash)[0]
+            f"{self.find_preps_ids_by_roothash(self.last_block.header.revealed_next_reps_hash)[0]}")
+        return self.find_preps_ids_by_roothash(self.last_block.header.revealed_next_reps_hash)[0]
 
     def get_expected_generator(self, peer_id: ExternalAddress) -> ExternalAddress:
         """get expected generator to vote unconfirmed block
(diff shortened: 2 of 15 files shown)


loopchain/blockchain/blocks/v0_3/block.py
```diff
@@ -25,6 +27,26 @@ class BlockHeader(BaseBlockHeader):
...
+            return PrepChangedReason.TERM_END
+
+        return PrepChangedReason.PENALTY
+
+    @property
+    def revealed_next_reps_hash(self):
+        if self.prep_changed:
+            return self.next_reps_hash
+        return self.reps_hash
+
```
metalg0su: those three method became used at both v0.1 and v0.3.
How about extract those things at `BaseBlockHeader`?

loopchain/blockchain/blocks/v0_3/block.py
```diff
@@ -25,6 +27,26 @@ class BlockHeader(BaseBlockHeader):
     def complained(self):
         return self.leader_votes_hash != Hash32.empty()
 
+    @property
+    def prep_changed(self) -> Optional[PrepChangedReason]:
```
metalg0su: I think its naming could raise confusion to return bool type (prep changed or not).
How about `prep_changed_reason` or...

as normal method: get_prep_changed_reason? I saw in your code below:
```python
            if self._block_manager.epoch.leader_id != ChannelProperty().peer_id \
                    and not candidate_block.header.prep_changed:
```
It could be:
```python
            if self._block_manager.epoch.leader_id != ChannelProperty().peer_id \
                    and not candidate_block.header.get_prep_changed_reason():
```

loopchain/blockchain/blocks/v0_3/block.py
```diff
@@ -25,6 +27,26 @@ class BlockHeader(BaseBlockHeader):
...
+        if self.next_reps_hash == Hash32.empty():
+            return None
+
+        if self.next_leader == ExternalAddress.empty():
+            return PrepChangedReason.TERM_END
+
+        return PrepChangedReason.PENALTY
+
+    @property
+    def revealed_next_reps_hash(self):
```
metalg0su: Can you explain `revealed` means?

loopchain/blockchain/blocks/v0_3/block_verifier.py
```diff
@@ -113,14 +115,22 @@ def verify_invoke(self, builder: 'BlockBuilder', block: 'Block', prev_block: 'Bl
...
         if header.next_reps_hash != new_block.header.next_reps_hash:
-            exception = RuntimeError(f"Block({header.height}, {header.hash.hex()}, "
-                                     f"NextRepsHash({header.next_reps_hash}), "
-                                     f"Expected({new_block.header.next_reps_hash}).")
-            self._handle_exception(exception)
+            if not new_block.header.prep_changed \
+                    and header.next_reps_hash == new_block.header.revealed_next_reps_hash:
+                pass
+            else:
+                exception = RuntimeError(f"Block({header.height}, {header.hash.hex()}, "
```
metalg0su: I think it is better to specify Exception Type for better legibility, like `PrepChangedButNoChangedInRepsHash` (--;)

loopchain/blockchain/blocks/v0_3/block.py
windies21: good idea!

windies21: But it's not simple. Because BaseBlockHeader is set to the data class. I postpone this to the next one.

loopchain/blockchain/blocks/v0_3/block.py
windies21: It will develop further to feature 'keep order'. In this PR it was used as a kind of boolean (of course It's not actual bool value).

yakkle (commented):
It would be better to read that `prep_changed` compare statements change like below.
```py
        if last_block.header.prep_changed == PrepChangedReason.PENALTY:
```

loopchain/baseservice/score_code.py
```diff
@@ -28,4 +28,4 @@ class ScoreResponse(IntEnum):
 
 class PrepChangedReason:
     TERM_END = "0x0"  # Term ends and reset P-Rep list
-    PANELTY = "0x1"  # Updated within term (Panelty)
+    PENALTY = "0x1"  # Updated within term (Penalty)
```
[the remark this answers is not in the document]
windies21: You can consider it for feature 'keep order' (LC-557). But It's not necessary in this PR.

loopchain/blockchain/blocks/v0_3/block.py
```diff
@@ -25,6 +27,26 @@ class BlockHeader(BaseBlockHeader):
...
+        if self.prep_changed:
```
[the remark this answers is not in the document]
windies21: Nope. Normal block has empty hash in its next_reps_hash but That's not true value. That means this block is normal block. So Engine can revealed next reps hash from that block. That block's real next_reps_hash is same to reps_hash.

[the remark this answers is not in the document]
windies21: We have to check if this block is normal. Penalty check is not good enough.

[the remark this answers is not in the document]
yakkle: When I see this line, I checked `prep_changed` function.
and check return value `None` or `PrepChangedReason.TERM_END` or `PrepChangedReason.PENALTY`.
Then, I have seen `PrepChangedReason` class to check real value of `TERM_END` and `PENALTY`.
at last, I got it `prep_changed` is True when `TERM_END`, `PENALTY`.

`prep_changed` function is not clear to understand `prep_changed` condition.
It disturb code reading flow.

So, I suggest like this for readability all over the `prep_changed` used.
```py
if self.prep_changed in (PrepChangedReason.TERM_END, PrepChangedReason.PENALTY):
```
or, change `prep_changed` function to `prep_changed_reason`.
and add `prep_changed` function.
```py
    def prep_changed(self) -> bool:
        return self.prep_changed_reason in (PrepChangedReason.TERM_END, PrepChangedReason.PENALTY):
```

loopchain/peer/consensus_siever.py
```diff
@@ -126,8 +126,8 @@ def __get_complaint_votes(self):
         util.logger.debug(f"-------------------consensus-------------------")
         async with self.__lock:
             if self._block_manager.epoch.leader_id != ChannelProperty().peer_id:
-                util.logger.warning(f"This peer is not leader. epoch leader={self._block_manager.epoch.leader_id}")
-                return
```
yakkle: why this return line removed?

loopchain/blockchain/blockchain.py
```diff
@@ -1139,9 +1137,9 @@ def score_invoke(self, _block: Block, prev_block: Block, is_block_editable: bool
             # P-Rep list has been changed
             utils.logger.debug(f"in score invoke current_height({_block.header.height}) next_prep({next_prep})")
 
-            if next_prep.state == PrepChangedReason.TERM_END:
-                next_leader = Hash32.empty()
-            elif next_prep.state == PrepChangedReason.PANELTY:
+            if next_prep["state"] == PrepChangedReason.TERM_END:
+                next_leader = ExternalAddress.empty()
+            elif next_prep["state"] == PrepChangedReason.PENALTY:
```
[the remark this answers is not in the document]
windies21: You can consider it for feature 'keep order' (LC-557). But It's not necessary in this PR.

loopchain/blockchain/blocks/v0_3/block_verifier.py
windies21: Not interested in this PR

loopchain/peer/consensus_siever.py
windies21: When the term transitions, the leader is not the leader but must make the last block. When leader make last block of Term, that leader is no more leader but must make the last block.

yakkle (approved):

loopchain/blockchain/blocks/v0_3/block_verifier.py
```diff
@@ -113,14 +115,22 @@ def verify_invoke(self, builder: 'BlockBuilder', block: 'Block', prev_block: 'Bl
...
             self._handle_exception(exception)
         if header.next_reps_hash != new_block.header.next_reps_hash:
-            exception = RuntimeError(f"Block({header.height}, {header.hash.hex()}, "
-                                     f"NextRepsHash({header.next_reps_hash}), "
-                                     f"Expected({new_block.header.next_reps_hash}).")
-            self._handle_exception(exception)
+            if not new_block.header.prep_changed \
+                    and header.next_reps_hash == new_block.header.revealed_next_reps_hash:
+                pass
+            else:
```
yakkle: I think `if (condition): else` statements can be replaced by the following condition.
```suggestion
            if (new_block.header.prep_changed or header.next_reps_hash != new_block.header.revealed_next_reps_hash):
```

windies21: Good suggestion.

yakkle:
retest this please

PR #319, opened 2019-09-16
[LC-588] Refactoring integration test

metalg0su:
# Goal
## Split out configuration maker
- Normalize making configuration for tests by `config_generator.py`
> `ConfigInterface` class aims to be used in integration tests
> Possible re-usage to generate temporary configurations to users.
> Test-able in `test_config_generator.py`

## Check process running in different way 
__No more pytest-xprocess!__
- Before) Check process running by waiting for specific regex pattern to appear.
- After) Ensure process run by using v1 avail API

## Increase readabilty
- For the clean world.
>  **__Note__**: Test Scenario is not so changed, but its shape has been changed many. So please review the current state of code, not the code of diff. Thank you


yakkle:
please add description

metalg0su:
Need some tools to generate config files.
Until then, close.

PR #336, opened 2019-09-20
[LC-618] Allow verify exception when normal block's next reps hash is

windies21:
Allow verify exception when normal block's next reps hash is empty (it is intended in v0.3 block.)

### DIFF
--- a/loopchain/blockchain/blocks/v0_3/block_verifier.py
+++ b/loopchain/blockchain/blocks/v0_3/block_verifier.py
@@ -116,14 +116,17 @@ class BlockVerifier(BaseBlockVerifier):
             self._handle_exception(exception)
 
         if header.next_reps_hash != new_block.header.next_reps_hash:
-            exception = RuntimeError(f"Block({header.height}, {header.hash.hex()}, "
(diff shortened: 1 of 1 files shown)


loopchain/blockchain/blocks/v0_3/block_verifier.py
```diff
@@ -116,14 +116,18 @@ def verify_invoke(self, builder: 'BlockBuilder', block: 'Block', prev_block: 'Bl
...
-                                     f"new block header({new_block.header}).")
-            self._handle_exception(exception)
+            if not new_block.header.prep_changed \
+                    and header.next_reps_hash == new_block.header.revealed_next_reps_hash:
+                pass
+            else:
+                exception = RuntimeError(f"Block({header.height}, {header.hash.hex()}, "
+                                         f"NextRepsHash({header.next_reps_hash}), "
+                                         f"Expected({new_block.header.next_reps_hash}), "
+                                         f"next_reps_hash({new_block.header.next_reps_hash}), "
```
hyerim-kim: It prints a same value with the upper line.

PR #341, opened 2019-09-24
[LC-619] set REST_TIMEOUT to 5 when select_fastest_endpoint

laurenceyoon:
- init_rs_target only when citizen node starts SubscribeNetwork
- reduce timeout to 5 when select fastest endpoint.
- if there's no alive endpoint when SubscribeNetwork, it follows the SUBSCRIBE_TIMER logic (retry 1 minute).


windies21 (approved):

loopchain/baseservice/rest_client.py
```diff
@@ -63,6 +64,21 @@ def _init_http_clients(self):
             if version != conf.ApiVersion.v1:
                 self._http_clients[url] = HTTPClient(url)
 
+    async def _fetch_status(self, session: ClientSession, request_uri):
+        endpoint_target = urlparse(request_uri).scheme + "://" + urlparse(request_uri).netloc
```
metalg0su: How about...:
```python
from urllib.parse import urljoin
...

endpoint_target = urljoin(request_uri, "/")
```

metalg0su: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin

laurenceyoon: it causes error when endpoint_target is set as urljoin because 443 port is added into dns url.

metalg0su (approved):

PR #343, opened 2019-09-24
[LC-625] Timer(Service) unit tests

metalg0su:
# Goal
- More precise test
> Test for callbacks both blocking and coroutine func.
- Faster Unit test for `Timer` & `TimerService`
> 21+ sec -> 2.0- sec
- Increase coverage for `Timer` & `TimerService`
> 79% -> 96%+


metalg0su:
retest this please

loopchain/baseservice/timer_service.py
```diff
@@ -66,9 +66,13 @@ def is_timeout(self) -> bool:
 
     @property
     def is_repeat(self) -> bool:
-        if self._is_repeat and \
-                (self._repeat_timeout == 0 or
-                 (time.time() - self.__repeat_start_time < self._repeat_timeout)):
+        is_repeatable = self._is_repeat
+        if not is_repeatable:
```
windies21: This temp variable is needed?

windies21 (approved):

loopchain/baseservice/timer_service.py
metalg0su: Thanks. You're right.

metalg0su: I removed 
FROM:
```
is_repeatable = self._is_repeat
if not is_repeatable:
```

TO:
```
+        if not self._is_repeat:
+            return False
```

PR #354, opened 2019-09-27
[LC-640] check UnrecordedBlock exception before vote it.

windies21:


loopchain/peer/block_manager.py
```diff
@@ -881,6 +888,8 @@ def verify_confirm_info(self, unconfirmed_block: Block):
         except InvalidUnconfirmedBlock as e:
             self.candidate_blocks.remove_block(unconfirmed_block.header.hash)
             util.logger.warning(e)
+        except UnrecordedBlock as e:
+            util.logger.info(e)
```
yakkle: exception -> logging info?

PR #372, opened 2019-10-15
[LC-690] apply eth_keyfile when loading prikey file

laurenceyoon:
To remove dependency of tbears, replaced it with eth_keyfile


metalg0su (approved):

windies21 (approved):

laurenceyoon:
I think `loop -r mainnet` should be working at the moment. At next release, they might be removed.

PR #392, opened 2019-10-24
[LC-719] Exclude block sync target if target_id is the block generator

laurenceyoon:
In case of block diversion, the voters / citizens would not sync blocks from leader node.

### DIFF
--- a/loopchain/blockchain/exception.py
+++ b/loopchain/blockchain/exception.py
@@ -34,6 +34,11 @@ class BlockError(Exception):
     pass
 
 
+class InvalidBlockSyncTarget(Exception):
+    """Raises when the target is invalid to block sync"""
+    pass
+
+
 class BlockchainError(Exception):
     """블럭체인상에서 문제가 발생했을때 발생하는 에러
     """
--- a/loopchain/peer/block_manager.py
+++ b/loopchain/peer/block_manager.py
@@ -19,7 +19,7 @@ import threading
 import traceback
 from collections import defaultdict
 from concurrent.futures import ThreadPoolExecutor, Future
-from typing import TYPE_CHECKING, Dict, DefaultDict, Optional, Tuple
+from typing import TYPE_CHECKING, Dict, DefaultDict, Optional, Tuple, List
 from pkg_resources import parse_version
 
 import loopchain.utils as util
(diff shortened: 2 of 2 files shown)


loopchain/peer/peer_outer_service.py
```diff
@@ -235,6 +235,9 @@ def GetStatus(self, request, context):
         if request.request == 'block_sync':
             try:
                 status_data = channel_stub.sync_task().get_status()
+                if (status_data['state'] == 'BlockGenerate') and (status_data['block_height'] > 1):
+                    status_data = None
+                    raise RuntimeError("Cannot accept BlockSync during BlockGenerate state")
```
metalg0su: Can you explain why did you set `status_data` as `None`, which raises `ChannelStatusError`?

laurenceyoon: That's because if status_data is not None, it (unnecessarily) caches the status_data. The purpose of this feature is to ignore block_sync request when the node gets leader. Because `StatusReply` doesn't have response_code as parameter, I couldn't send back to the requester with specific error code so let `ChannelStatusError` to be raised.

laurenceyoon: Changed again. It won't raise `ChannelStatusError` from Leader node. Instead, when the requester node send GetStatus(block_sync), it would recognize from status_data whether it's leader or not.

loopchain/peer/peer_outer_service.py
```diff
@@ -490,10 +490,22 @@ def AnnounceUnconfirmedBlock(self, request, context):
     def BlockSync(self, request, context):
         # Peer To Peer
         channel_name = conf.LOOPCHAIN_DEFAULT_CHANNEL if request.channel == '' else request.channel
+        channel_stub = StubCollection().channel_stubs[channel_name]
+        status_data = asyncio.run_coroutine_threadsafe(
+            channel_stub.async_task().get_status(),
+            self.peer_service.inner_service.loop
+        ).result()
+        if status_data["state"] == "BlockGenerate" and request.block_height > 1:
+            logging.debug(f"Cannot send BlockSync request.height({request.block_height})) during BlockGenerate")
```
metalg0su: Please consider: `Cannot reply to BlockSync...` or `Cannot send BlockSyncReply...`

windies21 (approved):

PR #401, opened 2019-11-06
[LC-728] Avoid synchronizing a block that doesn't have the `confirm_info` and next block.

hyerim-kim:
Avoid synchronizing a block that doesn't have the `confirm_info` and next block.

### DIFF
--- a/loopchain/peer/block_manager.py
+++ b/loopchain/peer/block_manager.py
@@ -20,18 +20,19 @@ import traceback
 from collections import defaultdict
 from concurrent.futures import ThreadPoolExecutor, Future
 from typing import TYPE_CHECKING, Dict, DefaultDict, Optional, Tuple, List
+
 from pkg_resources import parse_version
 
 import loopchain.utils as util
 from loopchain import configure as conf
 from loopchain.baseservice import TimerService, ObjectManager, Timer, RestMethod
 from loopchain.baseservice.aging_cache import AgingCache
-from loopchain.blockchain import BlockChain, CandidateBlocks, Epoch, BlockchainError, NID, exception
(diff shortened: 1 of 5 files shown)


loopchain/peer/block_manager.py
```diff
@@ -608,7 +610,7 @@ def __block_request_to_peers_in_sync(self, peer_stubs, my_height, unconfirmed_bl
                             util.exit_and_msg(f"This peer already tried to synchronize {my_height} block "
                                               f"for max retry number({conf.BLOCK_SYNC_RETRY_NUMBER}). "
                                               f"Peer will be down.")
-            else:
+            elif response_code == message_code.Response.fail:
```
laurenceyoon: Then when`response_code == message_code.Response.fail_no_confirm_info` case, shouldn't it retry sync to another target? I guess there's no changing of `peer_index` when case of no_confirm_info, so it might retry to same target again and again.

loopchain/peer/block_manager.py
```diff
@@ -406,23 +406,25 @@ def __block_request_by_voter(self, block_height, peer_stub):
...
-            raise exception.BlockError(f"Received block is invalid: original exception={e}")
 
-        votes_dumped: bytes = response.confirm_info
-        try:
-            votes_serialized = json.loads(votes_dumped)
-            votes = BlockVotes.deserialize_votes(votes_serialized)
-        except json.JSONDecodeError:
-            votes = votes_dumped
+        if response.response_code == message_code.Response.fail_no_confirm_info:
+            block = None
```
windies21: How about raise exception instead set None.

windies21 (approved):

laurenceyoon (approved):

PR #406, opened 2019-11-12
[LC-754] remove add_peer from channel_infos

laurenceyoon:
- In order to get initial rep list, the `crep_root_hash` should be in its configuration.
- `channel_infos` will contain only the `CHANNEL_OPTION` configuration, not peer list.

### DIFF
--- a/loopchain/channel/channel_service.py
+++ b/loopchain/channel/channel_service.py
@@ -417,14 +417,18 @@ class ChannelService:
                 else:
                     logging.warning(f"Waiting for next subscribe request...")
 
-        subscribe_event = asyncio.Event()
         utils.logger.spam(f"try subscribe_call_by_citizen target({ChannelProperty().rest_target})")
-
+        subscribe_event = asyncio.Event()
         # try websocket connection, and handle exception in callback
-        asyncio.ensure_future(self.__node_subscriber.start(
-            block_height=self.__block_manager.blockchain.block_height,
-            event=subscribe_event,
-        ), loop=MessageQueueService.loop).add_done_callback(_handle_exception)
+        task = asyncio.ensure_future(
+            self.__node_subscriber.start(
+                block_height=self.__block_manager.blockchain.block_height,
+                event=subscribe_event
+            ),
+            loop=MessageQueueService.loop
+        )
+        task.add_done_callback(_handle_exception)
+
         await subscribe_event.wait()
 
     def shutdown_peer(self, **kwargs):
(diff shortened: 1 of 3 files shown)


loopchain/channel/channel_service.py
```diff
@@ -417,14 +417,19 @@ def _handle_exception(future: asyncio.Future):
...
+            "block_height": self.__block_manager.blockchain.block_height,
+            "event": subscribe_event
+        }
         # try websocket connection, and handle exception in callback
-        asyncio.ensure_future(self.__node_subscriber.start(
-            block_height=self.__block_manager.blockchain.block_height,
-            event=subscribe_event,
-        ), loop=MessageQueueService.loop).add_done_callback(_handle_exception)
+        task = asyncio.ensure_future(
+            self.__node_subscriber.start(**subscribe_params),
```
metalg0su: How about pass `block_height` and `event` as a keyword args, not the unpacked dict?

loopchain/peermanager/peer_loader.py
```diff
@@ -49,13 +49,13 @@ def _load_peers_from_file(peer_manager: 'PeerManager'):
     @staticmethod
     def _load_peers_from_rest_call(peer_manager: 'PeerManager'):
         rs_client = ObjectManager().channel_service.rs_client
-        is_block_version_0_3 = PeerLoader._is_block_version_0_3(rs_client)
         crep_root_hash = conf.CHANNEL_OPTION[ChannelProperty().name].get('crep_root_hash')
 
-        if is_block_version_0_3 and crep_root_hash:
-            return PeerLoader._get_reps_by_root_hash_call(peer_manager, rs_client, crep_root_hash)
+        if not crep_root_hash:
+            logging.error(f"There's no crep_root_hash to initialize.")
```
metalg0su: The channel option `crep_root_hash` is not optional but required to run loopchain node, Right?

loopchain/peer/peer_service.py
```diff
@@ -281,4 +275,4 @@ def close(self):
             await StubCollection().create_icon_score_stub(channel_name)
 
     def _load_channel_infos(self):
-        self._channel_infos = self._get_channel_infos()
+        self._channel_infos = conf.CHANNEL_OPTION
```
metalg0su: Is this method needed?
How about replace `self._load_channel_infos()` as `self._channel_infos = conf.CHANNEL_OPTION`
and remove this method?

windies21 (approved):

loopchain/peer/peer_service.py
laurenceyoon: Good idea!

loopchain/peermanager/peer_loader.py
laurenceyoon: Yep. `crep_root_hash` is required for citizen node.

loopchain/channel/channel_service.py
laurenceyoon: Seems reasonable. Updated!

metalg0su (approved):

zsaladin (approved):

PR #420, opened 2019-11-25
[LC-755] Loopchain Configure Generator

metalg0su:
Try this:

`$ loop init --total-reps=8 --main-reps=4 --channel-names icon_dex loopchain_test --config-output=~/.loopchain`

### DIFF
--- a/loopchain/tools/config_gen/config_gen.py
+++ b/loopchain/tools/config_gen/config_gen.py
@@ -1,20 +1,20 @@
 import copy
 from pathlib import Path
-from typing import List, Iterable
+from typing import List
 
 from loopchain.tools.config_gen.helper import (
     check_param_exist, dict_write, make_port_by_order, make_radiostations_for_local,
     make_crep_root_hash, make_genesis_data, make_channel_config, make_peer_config, make_channel_manage_data
 )
 from loopchain.tools.config_gen.types import Key, Keys, Config
+from loopchain.tools.config_gen.const import *
 
 
 class ConfigGenerator:
-    def __init__(self, config_root: Path, channel_names: List[str], total_reps_count: int,
-                 main_reps_count: int = None,
-                 key_password: str = None, key_root: Path = None):
+    def __init__(self, config_root: Path, channel_names: List[str],
+                 total_reps_count: int, main_reps_count: int = None, key_password: str = None):
         # Required Inputs
-        self._config_root: Path = config_root.expanduser().resolve()
+        self._config_root = config_root.expanduser().resolve()
         self._total_reps_count: int = total_reps_count
         self._channel_names: List[str] = channel_names
 
@@ -27,8 +27,7 @@ class ConfigGenerator:
 
         # Optional
         self.main_reps_count: int = main_reps_count or total_reps_count
-        self.key_root: Path = key_root or self._config_root / "keys"
-        self.key_password: str = key_password or "password"
+        self.key_password: str = key_password or DEFAULT_KEY_PASSWORD
 
     def build(self):
         self.build_keys()
@@ -45,7 +44,7 @@ class ConfigGenerator:
 
         keys = []
         for i in range(self._total_reps_count):
-            key_path = self.key_root / f"my_keystore_{i}.json"
+            key_path = self._config_root / f"Node{i}" / DEFAULT_KEY_NAME
             key = Key(path=key_path, password=self.key_password)
             keys.append(key)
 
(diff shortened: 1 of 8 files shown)


loopchain/tools/config_gen/config_gen.py
```diff
@@ -0,0 +1,196 @@
...
+    def write_peer_configs(self) -> Iterable[Path]:
+        self._config_root.mkdir(exist_ok=True)
+
+        paths = []
+        for idx, peer_config in enumerate(self.peer_configs):
+            config_path = self._config_root / f"test_{idx}_conf.json"
+            paths.append(config_path)
+            dict_write(config_path, peer_config)
+
+        return paths
```
laurenceyoon: Why the return type of paths is different between `write_keys` and `write_peer_configs`? If it's intended, I think it needs to be applied in the tests as well.

loopchain/tools/config_gen/helper.py
```diff
@@ -0,0 +1,158 @@
...
+            else:
+                return func(self, *args, **kwargs)
+        return __check
+    return _check
+
+
+def dict_write(path: Union[Path, str], dict_obj: dict):
+    """Write dict value to path."""
+    if os.path.exists(path):
+        raise FileExistsError(f"{path} already exists.")
```
laurenceyoon: What if I want to regenerate config files with different number of total reps? Should I remove the directory before `loop init`?

testcase/unittest/tools/test_config_gen.py
```diff
@@ -0,0 +1,236 @@
...
+    def test_build_all_reps_key(self, config_gen_factory):
+        expected_total_reps = 8
+        config_gen = config_gen_factory(total_reps_count=expected_total_reps)
+        assert not config_gen.keys
+
+        config_gen.build_keys()
+        assert config_gen.keys
+        assert expected_total_reps == len(config_gen.keys)
+
+    def test_build_returns_its_data_if_exists(self, config_gen_factory):
```
laurenceyoon: What is the purpose of these methods below? Is it actually intended to set expected_data like test codes?

loopchain/tools/config_gen/helper.py
```diff
@@ -0,0 +1,158 @@
...
+        peers.append(peer)
+
+    return peers
+
+
+def make_channel_manage_data(channel_names: list, keys: Keys) -> Config:
+    channel_manage_data = {}
+    for channel_name in channel_names:
+        channel_manage_data[channel_name] = {
+            "score_package": "score/icx",  # TODO: Not needed?
```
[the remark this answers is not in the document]
metalg0su: Thank you :)

loopchain/tools/config_gen/config_gen.py
metalg0su: It's my mistake. I changed:
- return `List[Path]`, not Generator[Path]

testcase/unittest/tools/test_config_gen.py
metalg0su: I intended that `build_xxx` methods:
- build xxx if no data exists
- do nothing if xxx already exists as member.

Suppose that if someone wants to generate configure file with **exist key file**, not a generate new one. In this case, he or she is allowed to set his key file as `ConfigGenerator.keys`, then build rest of steps to get derived peer_configs etc.

loopchain/tools/config_gen/helper.py
metalg0su: Currently, yes. You have to remove configure root folder first.

---

Suppose that you generate 8 peers. Then you need to decrease the number of peer as 4. So you generate again 4 peers.
> In this case, if the config root folder exists, you will notice that [0-7]th peer configs are in root folder. [0-3] are new one, and [4-7] are old one.

I wanted to avoid this confusing situation. 

---

Do you any good idea??

loopchain/tools/config_gen/helper.py
```diff
@@ -0,0 +1,154 @@
...
+# ----- Channel Manage Data
+def _make_peers_in_channel_manage_data(keys: Keys) -> List[Config]:
+    peers = []
+
+    for order, key in enumerate(keys):
+        grpc_port = make_port_by_order(order, start_port=PORT_PEER)
+        peer = {
+            "id": key.address,
+            "peer_target": f"[local_ip]:{grpc_port}",
+            "order": order + 1
```
laurenceyoon: after merging LC-791, the "order" key is no more needed.

PR #428, opened 2019-12-02
[LC-768] Edit logic after `write_precommit_state`, which is determined by the result of itself.

hyerim-kim:
- Edit logic after `write_precommit_state`, which is determined by the result of itself.
- Add new exception class `WritePrecommitStateError`.

### DIFF
--- a/loopchain/channel/channel_service.py
+++ b/loopchain/channel/channel_service.py
@@ -25,7 +25,7 @@ from loopchain import utils
 from loopchain.baseservice import (BroadcastScheduler, BroadcastSchedulerFactory, BroadcastCommand,
                                    ObjectManager, CommonSubprocess, RestClient,
                                    NodeSubscriber, UnregisteredException, TimerService)
-from loopchain.blockchain.exception import AnnounceNewBlockError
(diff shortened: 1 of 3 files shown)


hyerim-kim:
retest this please

loopchain/channel/channel_service.py
```diff
@@ -553,7 +553,9 @@ def score_write_precommit_state(self, block: Block):
         request = convert_params(request, ParamType.write_precommit_state)
 
         stub = StubCollection().icon_score_stubs[ChannelProperty().name]
-        stub.sync_task().write_precommit_state(request)
+        precommit_result: dict = stub.sync_task().write_precommit_state(request)
+        if precommit_result != hex(0):
```
metalg0su: Since there are no public docs that shows the return value of icon-service score, `hex(0)` could raise a question. 

Please consider this:
```python
if "error" in precommit_result:
```

hyerim-kim:
retest this please

metalg0su (approved):

windies21 (approved):

PR #430, opened 2019-12-03
[LC-774] Add nid to '/api/v1/status/peer'.

windies21:
* Do caching peer status with lru_cache not status_cache in peer_service.
* Add nid to '/api/v1/status/peer'.

### DIFF
--- a/loopchain/peer/peer_outer_service.py
+++ b/loopchain/peer/peer_outer_service.py
@@ -1,29 +1,16 @@
-# Copyright 2018 ICON Foundation
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#     http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


loopchain/peer/peer_outer_service.py
```diff
@@ -181,22 +149,16 @@ def __get_status_peer_type_data(self, channel: str):
...
-                    return None
-            status_data = self.peer_service.status_cache[channel]
-        else:
-            channel_stub = StubCollection().channel_stubs[channel]
-            status_data = asyncio.run_coroutine_threadsafe(
-                channel_stub.async_task().get_status(),
-                self.peer_service.inner_service.loop
-            ).result()
-            self.peer_service.status_cache[channel] = status_data
+    @lru_cache(maxsize=4, valued_returns_only=True)
```
metalg0su: It stores status cache, but the only latest second signature is valid. (unless loopchain use `getStatus` to get data passed.)

Is there any reason for you to set `maxsize` as 4?
Every return value has large `dict`, contains many key-and-values.

I think 1 is enough.

windies21: Basically, the GetStatus return value is used for monitoring. (Except block sync). So 1 second delay is no problem in its purpose. And, I agree your opinion about maxsize of cache.

PR #439, opened 2019-12-13
[LC-821] update block sync logic for prevent deadlock during block_sync

laurenceyoon:
- fix bug in block_height_sync logic (for deadlock situation during block_sync)
- remove `role_switch_block_height` 
- add `stop_ws_heartbeat_timer`

### DIFF
--- a/loopchain/peer/block_manager.py
+++ b/loopchain/peer/block_manager.py
@@ -563,7 +563,7 @@ class BlockManager:
                 util.logger.warning(f"{e}")
                 response_code = message_code.Response.fail_no_confirm_info
             except Exception as e:
-                util.logger.warning("There is a bad peer, I hate you: " + str(e))
+                util.logger.warning(f"There is a bad peer, I hate you: {type(e), e}")
                 traceback.print_exc()
                 response_code = message_code.Response.fail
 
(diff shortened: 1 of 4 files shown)


loopchain/peer/block_manager.py
```diff
@@ -656,27 +639,16 @@ def _handle_exception(e):
...
-                                                                              my_height,
-                                                                              unconfirmed_block_height,
-                                                                              max_height)
+                self.__block_request_to_peers_in_sync(peer_stubs,
+                                                      my_height,
+                                                      unconfirmed_block_height,
+                                                      max_height)
         except Exception as exc:
             _handle_exception(exc)
             return False
```
metalg0su: Need to remove this return line, to ensure this func to end at the single point of truth.

laurenceyoon: No problem.

loopchain/peer/block_manager.py
```diff
@@ -578,50 +577,33 @@ def __block_request_to_peers_in_sync(self, peer_stubs, my_height, unconfirmed_bl
...
+                    elif self.blockchain.find_nid() is None:
+                        genesis_block = self.blockchain.find_block_by_height(0)
+                        self.__rebuild_nid(genesis_block)
 
                 except KeyError as e:
-                    result = False
-                    util.logger.error("fail block height sync: " + str(e))
-                    break
+                    util.logger.error(f"fail block height sync: {type(e), e}")
+                    raise
```
metalg0su: Do you know when this exception could be happened?

metalg0su: and.. I think more specific logging message needed. It looks very similar to log msg `except Exception as e` below

laurenceyoon: No idea for what triggers KeyError, but I've added traceback.

metalg0su (approved):

windies21 (approved):

PR #440, opened 2019-12-17
[LC-791] add load peers from DB

laurenceyoon:
- load `rep_root_hash` (as `last_block`'s `reps_hash` or `crep_root_hash` from the configuration) and find reps in DB if exists.

PR #444, opened 2019-12-18
[LC-707] add NodePool and RestClientProxy

laurenceyoon:
**[should be merged after LC-791(#440) is merged]**

- separate `NodePool` and add `RestClientProxy` from `RestClient` that manages nearest node endpoints


metalg0su:
Close Pending PR. Reopen when it needed.

PR #446, opened 2019-12-24
[LC-748] Add block data and vote data of version 5.

hyerim-kim:
- Add data of version 5 for remove `_` from a variable name of the `round_`.
- Add logic to get a class that is related to voting by version.
- Edit proto file of the gRPC.
- Replace some the `logging` to the `logger`.


loopchain/baseservice/node_subscriber.py
```diff
@@ -164,7 +164,8 @@ def __del__(self):
         block_dict, votes_dumped = kwargs.get('block'), kwargs.get('confirm_info', '')
         try:
             votes_serialized = json.loads(votes_dumped)
-            vote = BlockVotes.deserialize_votes(votes_serialized)
+            votes_class = utils.get_vote_class_by_version(block_dict["version"])["BlockVotes"]
```
windies21: This file no longer uses 'BlockVotes'. So import section is to be re-organized.

windies21 (approved):

PR #449, opened 2019-12-30
[LC-757] Do block rollback when exception PreviousBlockMismatch.

windies21:
- Do block rollback when exception PreviousBlockMismatch.

- Remove KI0 blocks when rollback.

- check_rollback_possible before roll back.


loopchain/blockchain/blockchain.py
```diff
@@ -227,6 +224,67 @@ def close_blockchain_store(self):
...
+    def check_rollback_possible(self, target_block, start_block=None):
+        """Check if the target block can be reached with the prev_hash of last_block.
+        CAUTION! This method is called recursively.
+
+        :param target_block:
+        :param start_block:
+        :return:
+        """
+        if not start_block:
+            start_block = self.__last_block
```
yourhrh: self.__last_block is not thread safe. if blockchain is in competition state in multi thread, call with same args but result is different. i think this is bad pattern. i suggest this function is always called with start_block arg

PR #451, opened 2020-01-02
Add ICON badge

JINWOO-J:
Added an ICON badge in README.md
Fix it again in the develop branch

PR #437, opened 2019-12-10
[LC-765] Fix logic to figure out the result of a leader complaint.

hyerim-kim:
Fix logic to figure out the result of a leader complaint.
- Edit and add some tests.
- Remove logic to process the vote of `out of round` in `add_complain`.
- Remove some logs.

PR #453, opened 2020-01-03
Merge to develop from release/2.5.0

metalg0su:
# Changes
- LC-821: update block sync logic for prevent deadlock during block_sync #439
- LC-765: Fix logic to figure out the result of a leader complaint. #437
- LC-748: Add block data and vote data of version 5. #446
- LC-832: move subscribe event setting code to _handshake #445
- LC-834: show version of loopchain and it's dependencies #447
- LC-757: Do block rollback when exception PreviousBlockMismatch. #449
- Add ICON badge #451


Makefile
```diff
@@ -42,7 +42,7 @@ all: install generate-key
 requires:
 	$(PIP_INSTALL) iconservice==1.5.15
```
metalg0su: NEED TO BE CHANGED

windies21 (approved):

PR #454, opened 2020-01-07
[LC-806] Add logic to reset network.

hyerim-kim:
(coupled: https://github.com/icon-project/icon-rpc-server/pull/124)

Add logic to reset network when a node gets a request that wants to send a block from a subscriber that has a higher height than it.
- Add a transition `reset_network` in the `state machine` of the channel.
- Replace `logging` to `utils.logger`.
- Edit some logs.
- Remove unnecessary logs.


yakkle:
stale

PR #455, opened 2020-01-07
[LC-829] Fix the problem to connect just one radiostation targets.

hyerim-kim:
Fix the problem to connect just one radiostation targets even if the node has more targets while the network is initialized.

### DIFF
--- a/loopchain/peer/block_manager.py
+++ b/loopchain/peer/block_manager.py
@@ -1,5 +1,5 @@
 """A management class for blockchain."""
-
+import asyncio
 import json
 import threading
 import traceback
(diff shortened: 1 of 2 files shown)


loopchain/peer/block_manager.py
```diff
@@ -430,21 +429,24 @@ def __block_request_by_voter(self, block_height, peer_stub):
...
+    async def __block_request_by_citizen(self, block_height):
         rs_client = ObjectManager().channel_service.rs_client
         get_block_result = rs_client.call(
             RestMethod.GetBlockByHeight,
             RestMethod.GetBlockByHeight.value.params(height=str(block_height))
         )
         last_block = rs_client.call(RestMethod.GetLastBlock)
+
         if not last_block:
+            await self.__channel_service.init_rs_target()
```
windies21: How about doing this in an exception handler? (Just my opinion)

yakkle:
stale

PR #459, opened 2020-01-09
[LC-845] PeerLoader.load () only when needed.

windies21:


loopchain/blockchain/blockchain.py
```diff
@@ -98,6 +99,8 @@ def __init__(self, channel_name=None, store_id=None, block_manager=None):
         for tx_version, tx_hash_version in channel_option.get("hash_versions", {}).items():
             self.__tx_versioner.hash_generator_versions[tx_version] = tx_hash_version
 
+        self._init_blockchain()
```
metalg0su: I like this line!
Good for init object when initialized, not in the other classes! 
It helps increasing readability I think.

loopchain/channel/channel_service.py
```diff
@@ -467,8 +464,6 @@ def _is_genesis_node(self):
                 and self.is_support_node_function(conf.NodeFunction.Vote))
 
     def __ready_to_height_sync(self):
-        self.block_manager.blockchain.init_blockchain()
-
```
metalg0su: https://github.com/icon-project/loopchain/pull/459/commits/bf5fa2807bf9e4170e97d64ed7efee56f0ff6c94#r364535934

loopchain/blockchain/blockchain.py
```diff
@@ -523,6 +526,15 @@ def find_preps_by_roothash(self, roothash: Hash32) -> list:
...
 
+    @valued_only_lru_cache(maxsize=4, valued_returns_only=True)
+    def roothash_exist_in_db(self, roothash: Hash32) -> Optional[bool]:
+        try:
+            self._blockchain_store.get(BlockChain.PREPS_KEY + roothash)
+        except (KeyError, TypeError):
+            return None
+        else:
+            return True
+
```
metalg0su: Did you mean:
```python
    @valued_only_lru_cache(maxsize=4, valued_returns_only=True)
    def roothash_exist_in_db(self, roothash: Hash32) -> Optional[bool]:
        return self.find_preps_by_roothash(roothash)
```

metalg0su: If it is close to your original intention, It is better for `roothash_exist_in_db` to follow `find_preps_by_roothash`'s return value.

loopchain/peermanager/peer_manager.py
```diff
@@ -38,11 +26,11 @@ def __init__(self):
         # reps_hash, reps for reset_all_peers
         self._reps_reset_data: Optional[tuple] = None
 
-        self._prepared_reps_hash = None
+        self._crep_root_hash = Hash32.fromhex(conf.CHANNEL_OPTION[ChannelProperty().name].get('crep_root_hash'))
```
metalg0su: Suggestion:

I think `PeerManager` needs to channel_name to be initialized. Please consider:
```python
class PeerManager:
    def __init__(self, channel_name):
        ....
        self._crep_root_hash = Hash32.fromhex(conf.CHANNEL_OPTION[channel_name].get('crep_root_hash'))
```
Using Singleton property to initialize `PeerManager` is risky.

I checked that `ChannelService` set its name ahead of `PeerManager` init.
But someday if changed initialization order, which means PeerManager first then ChannelProperty follows, it could be hard to find which line raises error.

loopchain/peermanager/peer_manager.py
```diff
@@ -59,16 +47,15 @@ def serialize_as_preps(self) -> list:
...
-        reps_in_db = blockchain.find_preps_by_roothash(reps_hash)
-
-        if not reps_in_db:
-            preps = self.serialize_as_preps()
-            util.logger.spam(f"in _load_peers serialize_as_preps({preps})")
-            blockchain.write_preps(reps_hash, preps)
+        if not blockchain.roothash_exist_in_db(self._crep_root_hash):
+            PeerLoader.load(peer_manager=self)
+            reps_hash = self.reps_hash()
+            reps_in_db = blockchain.roothash_exist_in_db(reps_hash)
```
metalg0su: Expected different return value caused by this line. `roothash_exist_in_db` returns boolean.

https://github.com/icon-project/loopchain/pull/459/files#r364537455

loopchain/blockchain/blockchain.py
windies21: I think 'is_' prefix is ​​not required because it can be identified by type hint. But I will accept your opinion.

loopchain/channel/channel_service.py
windies21: ??

loopchain/peermanager/peer_manager.py
windies21: I understand your worries. But this is not the last form. Peer Manager will be modified to not handle this 'crep_root_hash' in subsequent refactorings.

PR #461, opened 2020-01-13
[LC-272] fixed to add limited tx_items to queue

yakkle:
- tx_items has maximum items that not over tx_max_size or tx_max_count
 - monkeypatch configure value in test_custom_tx_value
 - fixed blocking issue when set maxsize in TxMessagesQueue

## current
`tx_list` added until over limited size or count from queue.
First take it from the queue, then check the status, and then add it back to the queue if the condition is false. So, re-added item is located on last of queue.

## will be
Add `tx_item` to a limited size or count. `tx_list` is now readily available from the queue.


testcase/unittest/test_tx_message.py
```diff
@@ -0,0 +1,149 @@
...
+    tx_size = 0
+
+    @pytest.mark.parametrize("tx_items", [tx_count], indirect=True)
+    def test_tx_messages(self, tx_items):
+        # given
+        self.tx_messages = TxMessages()
+        for tx_param in tx_items:
+            tx_item = TxItem.create_tx_item(tx_param, 'icon_dex')
+            self.tx_messages.append(tx_item)
+            self.tx_size += len(tx_item)
```
metalg0su: Any reason to define `tx_size` as class variable? 
It looks only being used inside this test.

testcase/unittest/test_tx_message.py
```diff
@@ -0,0 +1,149 @@
...
+from loopchain.baseservice import TxMessagesQueue, TxItem
+from loopchain.baseservice.tx_message import TxMessages
+from loopchain.blockchain import ExternalAddress
+from loopchain.blockchain.transactions import TransactionVersioner, TransactionBuilder
+from loopchain.crypto.signature import Signer
+
+
+@pytest.fixture
+def tx_item():
+    def _tx_item(tx_versioner):
```
metalg0su: Please consider to add type hints.
It helps understanding test cases.

```python
def tx_item() -> Callable(..., [Transaction]):
    def _tx_item(tx_versioner) -> Transaction:
        ...
```

testcase/unittest/test_tx_message.py
yakkle: ok. I will fix it.

loopchain/baseservice/tx_message.py
```diff
@@ -0,0 +1,106 @@
...
+        tx_total_size = self._tx_messages.total_size + len(tx_item)
+        tx_total_count = len(self._tx_messages) + 1
+
+        if tx_total_size >= self.max_tx_size or tx_total_count >= self.max_tx_count:
+            self._queue.put(self._tx_messages)
+            self._tx_messages = TxMessages()
+
+        self._tx_messages.append(tx_item)
+
+    def remain(self) -> bool:
```
metalg0su: I suggest:
```python
def is_empty(self) -> bool:
```
OR consider to follow `queue` module convention...:
```python
def empty(self) -> bool:
```

loopchain/baseservice/tx_message.py
```diff
@@ -0,0 +1,106 @@
...
+        return not self._queue.empty() or len(self._tx_messages) > 0
+
+    def size(self) -> int:
+        return self._queue.qsize()
+
+    def __str__(self):
+        return (f"{self.__class__.__name__}(queue={self._queue.qsize()}, "
+                f"tx_total_count={len(self._tx_messages)})")
```
metalg0su: Any reason to place `__str__` magic method to bottom of class, unlike other classes (`TxItem`, `TxMessages`)?

loopchain/baseservice/tx_message.py
```diff
@@ -0,0 +1,106 @@
...
+    def append(self, tx_item):
+        self._transactions.append(tx_item)
+        self.total_size += len(tx_item)
+
+    def get_messages(self) -> List:
+        tx_messages = [tx.get_tx_message() for tx in self._transactions]
+
+        return tx_messages
+
+    def reset(self):
```
metalg0su: How about naming `clear`, to follow `list` protocol in python?

testcase/unittest/test_tx_message.py
```diff
@@ -0,0 +1,166 @@
...
+    tx_count = 50
+
+    def test_custom_tx_value(self):
+        print()
+        # given custom max_tx_size and max_tx_count
+        given_max_tx_size = 1024
+        given_max_tx_count = 100
+
+        configure.MAX_TX_SIZE_IN_BLOCK = given_max_tx_size
+        configure.MAX_TX_COUNT_IN_ADDTX_LIST = given_max_tx_count
```
metalg0su: Please consider monkeypatch global variables: 
```python
        mocker.patch.object(configure, "MAX_TX_SIZE_IN_BLOCK", given_max_tx_size)
        mocker.patch.object(configure, "MAX_TX_COUNT_IN_ADDTX_LIST", given_max_tx_count)
```

You try to change global variables. It might be effective to other tests which uses those variables.
Try to print them out in another tests if they are changed or not.

loopchain/baseservice/tx_message.py
```diff
@@ -0,0 +1,106 @@
...
+
+        return tx_messages
+
+    def append(self, tx_item: TxItem):
+        tx_total_size = self._tx_messages.total_size + len(tx_item)
+        tx_total_count = len(self._tx_messages) + 1
+
+        if tx_total_size >= self.max_tx_size or tx_total_count >= self.max_tx_count:
+            self._queue.put(self._tx_messages)
+            self._tx_messages = TxMessages()
```
metalg0su: Possible violation of thread safeness.

The method might put `self._tx_messages` into `self._queue` **BEFORE** `TxMessages` initialized (cleared) in method `get`.

metalg0su: See those lines in `TxMessagesQueue.get`:
```python
    def get(self) -> Optional[TxMessages]:
            ...
            tx_messages = self._tx_messages
            if len(tx_messages) == 0:
                return None
            self._tx_messages = TxMessages()
            ...
```

testcase/unittest/test_tx_message.py
```diff
@@ -0,0 +1,166 @@
...
+        def consumer(a_count):
+            while True:
+                tx_messages = queue.get()
+                a_count['consume_count'] += len(tx_messages)
+                message = tx_messages.get_messages()
+                print(f"consumer() : queue = {queue}, tx_messages({id(tx_messages)})")
+                print(f"consume_count : {a_count['consume_count']}")
+                if not queue.remain():
+                    break
+                time.sleep(0.3)
```
metalg0su: Consider to reduce loop interval time.
If: 
- `producer` can 300 `TxItem`s per consumer loop **AND**
- `max_tx_count` of `queue` is set 21
then:
- This test **Always** get `TxMessages` from `TxMessagesQueue._queue`, not from `TxMessagesQueue._tx_messages` which might break thread safeness.

testcase/unittest/test_tx_message.py
```diff
@@ -0,0 +1,166 @@
...
+        # print(f"tx : {tx.raw_data}")
+        tx_params.append((tx, tx_versioner))
+
+    return tx_params
+
+
+class TestTxMessages:
+    tx_count = 50
+
+    @pytest.mark.parametrize("tx_items", [tx_count], indirect=True)
```
yakkle: ref : https://stackoverflow.com/a/33879151

loopchain/baseservice/tx_message.py
yakkle: No reason. Is there guide that recommends the location of a magic method like `PEP8`?

testcase/unittest/test_tx_message.py
yakkle: updated

loopchain/baseservice/tx_message.py
yakkle: updated

loopchain/baseservice/tx_message.py
yakkle: removed

testcase/unittest/test_tx_message.py
metalg0su: SUPER USEFUL parametrized fixture!

loopchain/baseservice/tx_message.py
```diff
@@ -0,0 +1,106 @@
...
+    def get_messages(self) -> List['TxSend']:
+        return self._messages
+
+    def size(self) -> int:
+        return self._total_size
+
+
+class TxMessagesQueue(queue.Queue):
+    def __init__(self, maxsize=0, max_tx_size=None, max_tx_count=None):
+        super().__init__(maxsize=maxsize)
```
metalg0su: `maxsize` of `queue.Queue` may lead the thread stuck at `self.put`.

Suppose that `maxsize` is supplied to `TxMessagesQueue`. And it reaches to its `maxsize` by appending tx_items.

In this case, `queue.Queue.put` will be blocked until the item of `queue.Queue` is consumed.
But this blocking situation cannot be resolved because `append` has locked and no other threads can `get` to consume the queue. 

This means that the running thread is blocked forever.

metalg0su (approved):

PR #472, opened 2020-02-05
[LC-873] Modified structure for new consensus

zsaladin:
When `ConsensusChanged` is raised new consensus has to work.

### DIFF
--- a/loopchain/app/app.py
+++ b/loopchain/app/app.py
@@ -1,0 +1,7 @@
+import logging
+
+
+class App:
+    def __init__(self):
+        logging.info("Initialized")
+        exit()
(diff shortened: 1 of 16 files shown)


loopchain/app/app.py
```diff
@@ -0,0 +1,7 @@
+import logging
+
+
+class App:
```
yourhrh: Do you use 'App' for name of new engine?
This name is written by viewpoint in LFT library but we need another name for loopchain.
I suggest new name 'BlockchainService' or 'LoopchainService_v3'.
How do you think about this name?

zsaladin: This PR is for separating loopchain into new structure and old structure.
We have to decide the name on next PR.

PR #473, opened 2020-02-07
[LC-382] replace package secp256k1 with coincurve

yakkle:
[secp256k1](https://github.com/ludbb/secp256k1-py) is no longer managed and replaced by [coincurve](https://github.com/ofek/coincurve).
- replace package secp256k1 with coincurve
- update icx_wallet to use coincurve

### DIFF
--- a/loopchain/crypto/signature.py
+++ b/loopchain/crypto/signature.py
@@ -1,26 +1,10 @@
-# Copyright 2018 ICON Foundation
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#     http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-""" A class for icx authorization of Peer"""
-
 import binascii
 import hashlib
 import logging
 from typing import Union, Type, TypeVar
 
 import eth_keyfile
-from secp256k1 import Base, ALL_FLAGS
-from secp256k1 import PrivateKey, PublicKey
+from coincurve import PrivateKey, PublicKey
 
 from loopchain.crypto.cert_serializers import DerSerializer, PemSerializer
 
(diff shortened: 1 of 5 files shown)


yourhrh (commented):
Need Interoperability test about secp256k1 signature blocks

loopchain/crypto/signature.py
```diff
@@ -141,11 +132,8 @@ def sign(self, data, is_hash: bool):
             logging.error(f"data must be bytes \n")
             return None
 
-        raw_sig = self.private_key.ecdsa_sign_recoverable(msg=data,
-                                                          raw=is_hash,
-                                                          digest=hashlib.sha3_256)
-        serialized_sig, recover_id = self.private_key.ecdsa_recoverable_serialize(raw_sig)
-        return serialized_sig + bytes((recover_id, ))
```
yakkle: These parts are different between `secp256k1-py` and `coincurve`.

### first, recovery id is one of [0, 1, 2, 3]

[here1](https://bitcoin.stackexchange.com/a/38909)
> That number between 0 and 3 we call the recovery id

[here2](https://github.com/bitcoin-core/secp256k1/blob/master/include/secp256k1_recovery.h#L34)
> recid:   the recovery id (0, 1, 2 or 3)

### second, `coincurve` call `int_to_bytes` with `recid` parameter
[serialize_recoverable](https://github.com/ofek/coincurve/blob/12.0.0/coincurve/ecdsa.py#L46)
```py
    return bytes(ffi.buffer(output, CDATA_SIG_LENGTH)) + int_to_bytes(recid[0])
```
[int_to_bytes](https://github.com/ofek/coincurve/blob/12.0.0/coincurve/utils.py#L45)
```py
    def int_to_bytes(num):
        return num.to_bytes((num.bit_length() + 7) // 8 or 1, 'big')
```

num is `recid`, and `int_to_bytes` return `num.to_bytes(1, 'big')` because of
```py
[ins] In [461]: [(i.bit_length()+7) // 8 or 1 for i in range(4)]
Out[461]: [1, 1, 1, 1]
```
So, `int_to_bytes` return always bytes of length 1
```py
[ins] In [463]: [i.to_bytes(1, 'big') for i in range(4)]
Out[463]: [b'\x00', b'\x01', b'\x02', b'\x03']
```

### finally, It is equivalent to `bytes((recid,))`
```py
[ins] In [464]: [bytes((i,)) for i in range(4)]
Out[464]: [b'\x00', b'\x01', b'\x02', b'\x03']
```

yourhrh: good

yourhrh (approved):

PR #476, opened 2020-02-24
[LC-891] Update jsonrpcclient version

metalg0su:
(coupled: https://github.com/icon-project/icon-rpc-server/pull/128)

# Goal
Update version of jsonrpcclient library from 2.6.0 to 3.3.5 (latest)

### DIFF
--- a/loopchain/baseservice/rest_client.py
+++ b/loopchain/baseservice/rest_client.py
@@ -23,10 +23,11 @@ from urllib.parse import urlparse
 
 import requests
 from aiohttp import ClientSession
-from jsonrpcclient import HTTPClient, Request
-from jsonrpcclient.aiohttp_client import aiohttpClient
-from loopchain import utils, configure as conf
+from jsonrpcclient.clients.aiohttp_client import AiohttpClient
+from jsonrpcclient.clients.http_client import HTTPClient
+from jsonrpcclient.requests import Request
 
+from loopchain import utils, configure as conf
 
 _RestMethod = namedtuple("_RestMethod", "version name params")
 
@@ -168,7 +169,7 @@ class RestClient:
         # 'aioHttpClient' does not support 'timeout'
         url = self._create_jsonrpc_url(target, method)
         async with ClientSession() as session:
-            http_client = aiohttpClient(session, url)
+            http_client = AiohttpClient(session, url)
             request = self._create_jsonrpc_params(method, params)
             return await http_client.send(request)
 
--- a/loopchain/baseservice/rest_client.py
+++ b/loopchain/baseservice/rest_client.py
@@ -165,11 +165,10 @@ class RestClient:
                                    timeout=timeout) as response:
                 return await response.json()
 
-    async def _call_async_jsonrpc(self, target: str, method: RestMethod, params: Optional[NamedTuple], timeout):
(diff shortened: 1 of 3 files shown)


loopchain/baseservice/rest_client.py
```diff
@@ -168,7 +169,7 @@ def _call_jsonrpc(self, target: str, method: RestMethod, params: Optional[NamedT
         # 'aioHttpClient' does not support 'timeout'
         url = self._create_jsonrpc_url(target, method)
         async with ClientSession() as session:
-            http_client = aiohttpClient(session, url)
+            http_client = AiohttpClient(session, url)
```
zsaladin: Can you put `timeout` as a parameter?
As you can see an comment above, the old version JsonRPCClient does not support timeout parameter. But this one seems to support it.

loopchain/baseservice/rest_client.py
```diff
@@ -206,5 +206,5 @@ def _create_jsonrpc_params(self, method: RestMethod, params: Optional[NamedTuple
             if "from_" in params:
                 params["from"] = params.pop("from_")
 
-        return Request(method.value.name, params) if params else Request(method.value.name)
+        return Request(method.value.name, **params) if params else Request(method.value.name)
```
yakkle: I wonder if this change does the same thing.
param match to `args` in `Request(method.value.name, params)`,
param match to `kwargs` in `Request(method.value.name, **params)`.

loopchain/baseservice/rest_client.py
metalg0su: Added. 
Please check this commit and resolve if it is acceptable.

https://github.com/icon-project/loopchain/pull/476/commits/540346d7ad68e846819c2dd14d57930158f7ce5d

loopchain/baseservice/rest_client.py
metalg0su: Without unpacking params, it will causes unexpected situation.
AS `params` treated as one positional argument, `Request` makes `params` as a `list` (`{'params': [{'key': 'value'}]}`), not a desired `dict` type (`{'params': {'key': 'value'}}`)

Try to omit the commit `54a60e` in this branch and run test. (`$ pytest testcase/unittest/test_rest_client.py`)

See: https://github.com/bcb/jsonrpcclient/commit/38b90394311e27624b86f6e156b0b3e42e97f535

yakkle: how does work well in `jsonrpcclient 2.6.0`?

zsaladin: @yakkle 
There have been many changes among the major update of JsonRPCClient.

2.6.0
https://github.com/bcb/jsonrpcclient/blob/2.6.0/jsonrpcclient/request.py#L99
If the parameter has only one dict, then it will be treated as keyword parameters.

3.4.0
https://github.com/bcb/jsonrpcclient/blob/3.3.4/jsonrpcclient/requests.py#L105
The codes have been removed.

zsaladin (approved):

loopchain/baseservice/rest_client.py
yakkle: ok. It looks like using bug in `2.6.0`.

yakkle (approved):

PR #480, opened 2020-03-02
[LC-891] Fix mismatch in (Aio)HttpClient.send Protocol

metalg0su:
(coupled: https://github.com/icon-project/icon-rpc-server/pull/130)

# Please read
- `jsonrpcclient.client.Client.send` ensures `jsonrpcclient.response.Response`.
- `jsonrpcclient.response.Response` has its response result in `.data.result`  
> See: https://jsonrpcclient.readthedocs.io/en/latest/api.html#response

# Goal
- Fix violation of API above. (Caused by: https://github.com/icon-project/loopchain/pull/476)

### DIFF
--- a/loopchain/baseservice/rest_client.py
+++ b/loopchain/baseservice/rest_client.py
@@ -26,7 +26,7 @@ from aiohttp import ClientSession
 from jsonrpcclient.clients.aiohttp_client import AiohttpClient
 from jsonrpcclient.clients.http_client import HTTPClient
 from jsonrpcclient.requests import Request
-
+from jsonrpcclient.response import Response
 from loopchain import utils, configure as conf
 
 _RestMethod = namedtuple("_RestMethod", "version name params")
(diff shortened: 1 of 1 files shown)


loopchain/baseservice/rest_client.py
```diff
@@ -154,7 +154,9 @@ def _call_jsonrpc(self, target: str, method: RestMethod, params: Optional[NamedT
         url = self._create_jsonrpc_url(target, method)
         http_client = HTTPClient(url)
         request = self._create_jsonrpc_params(method, params)
-        return http_client.send(request, timeout=timeout)
+
+        response: Response = http_client.send(request, timeout=timeout)
+        return response.data.result
```
zsaladin: What if response or response.data is None?

metalg0su: Thank you for your precious review.
I have been tested in various situations and reached to an conclusion:
- `Response.data` is ensured if remote call is in success.

If failed by calling unknown method or exception raised in server side, client raises `ReceivedNon2xxResponseError`, which contains error code (404 or 500, ...)

zsaladin (approved):

loopchain/baseservice/rest_client.py
yakkle: `response.data` is `Union[JSONRPCResponse, List[JSONRPCResponse]]` type and has `result` instance variable in SuccessResponse.
It would be better to process when single response and assert `response.data` is SuccessResponse.

metalg0su: how about resolve risks of batch responses by:

```py
        response: Response = http_client.send(request, timeout=timeout)
        if isinstance(response.data, list):
            raise NotImplementedError(f"Received batch response. Data: {response.data}")
        else:
            return response.data.result
```

yakkle (approved):

PR #481, opened 2020-03-03
[LC-896] Add LFT `app` to loopchain

zsaladin:
`app` of LFT is added to loopchain.
The integration with loopchain means modifying the `app` to work well with components of loopchain.


loopchain/launcher.py
```diff
@@ -116,7 +118,14 @@ def start_as_channel(args):
...
 
-    ChannelService(channel, amqp_target, amqp_key).serve()
+    try:
+        ChannelService(channel, amqp_target, amqp_key).serve()
+    except ConsensusChanged as e:
+        utils.logger.info(f"Consensus Changed")
+        utils.logger.info(f"Remain txs. {len(e.remain_txs)}")
+        utils.logger.info(f"Last unconfirmed block {e.last_unconfirmed_block.header}")
+        utils.logger.info(f"Last unconfirmed votes {e.last_unconfirmed_votes}")
+        Node(e.node_id).start()
```
yourhrh: I think new node needs all informations of ConsensusChanged exception. Will this pass in other feature?

zsaladin: Yeah, Purpose of this PR is just for working without any problem with minimum number of modifications from original LFT `app`.

yourhrh (approved):

loopchain/channel/channel_service.py
```diff
@@ -466,8 +466,8 @@ def __ready_to_height_sync(self):
             if self._is_genesis_node():
                 self.generate_genesis_block()
 
-        if not self.is_support_node_function(conf.NodeFunction.Vote) and not ChannelProperty().rs_target:
-            utils.exit_and_msg(f"There's no radiostation target to sync block.")
+        # if not self.is_support_node_function(conf.NodeFunction.Vote) and not ChannelProperty().rs_target:
+        #     utils.exit_and_msg(f"There's no radiostation target to sync block.")
```
yakkle: Why don't you remove if this lines useless? or add short description why you add comment this lines.

loopchain/app/ui/console.py
```diff
@@ -0,0 +1,52 @@
+import asyncio
+import IPython
```
yakkle: Is `IPtyhon` library essential on running loopchain?
What is the purpose of the `ui.console`?

zsaladin: The console is a part of LFT application. It is for easy debugging and operate node's action in runtime.
It is like web3.js or web3.py in ethereum.

It has to be installed in develop mode.
But now we will use LFT's console itself.

loopchain/channel/channel_service.py
zsaladin: The PR is for operating loopchain as a single node. So RS Target is not required. But current loopchain requires RS Target. 
The reason why I leave it as a commit is We do not decide how we solve it yet.

metalg0su:
retest this please

PR #482, opened 2020-03-04
[LC-897] Version up jsonrpcserver (3.5.6 -> 4.1.2)

metalg0su:
(coupled: https://github.com/icon-project/icon-rpc-server/pull/131)

# Changes
- `GenericJsonRpcServerError` now inherits `jsonrpcserver.exceptions.ApiError` (`JsonRpcServerError` deprecated in `jsonrpcserver`)
- `jsonrpcserver.config` is no more used (also deprecated in `jsonrpcserver`).
- Change dispatcher param type. It needs dumped string, not a dict itself.
- Add dispatched methods to `jsonrpcserver.methods.Methods`, not to `AsyncMethods` (`AsyncMethods` deprecated).

### DIFF
--- a/loopchain/baseservice/node_subscriber.py
+++ b/loopchain/baseservice/node_subscriber.py
@@ -22,7 +22,6 @@ from urllib import parse
 import websockets
 from earlgrey import MessageQueueService
 from jsonrpcclient.requests import Request
-from jsonrpcserver import config
 from jsonrpcserver.aio import AsyncMethods
 from websockets import WebSocketClientProtocol
 
@@ -35,8 +34,6 @@ from loopchain.blockchain.votes import Votes
 from loopchain.channel.channel_property import ChannelProperty
 from loopchain.protos import message_code
 
-config.log_requests = False
-config.log_responses = False
 ws_methods = AsyncMethods()
 CONNECTION_FAIL_CONDITIONS = {
     message_code.Response.fail_subscribe_limit,
--- a/loopchain/baseservice/node_subscriber.py
+++ b/loopchain/baseservice/node_subscriber.py
@@ -22,7 +22,8 @@ from urllib import parse
 import websockets
 from earlgrey import MessageQueueService
 from jsonrpcclient.requests import Request
-from jsonrpcserver.aio import AsyncMethods
+from jsonrpcserver import async_dispatch
+from jsonrpcserver.methods import Methods
 from websockets import WebSocketClientProtocol
 
 from loopchain import configure as conf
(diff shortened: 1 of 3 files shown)


zsaladin (approved):

loopchain/baseservice/node_subscriber.py
```diff
@@ -132,13 +134,13 @@ def __del__(self):
...
-        response: bytes = await asyncio.wait_for(
+        response: str = await asyncio.wait_for(
             fut=self._websocket.recv(),
             timeout=2 * conf.TIMEOUT_FOR_WS_HEARTBEAT
         )
-        response_dict = convert_response_to_dict(response)
+        response = validate_response(response)
 
-        await ws_methods.dispatch(response_dict)
+        await async_dispatch(response, ws_methods)
```
yakkle: `methods` parameter not passed, `dispatch` use global_methods.
https://github.com/bcb/jsonrpcserver/blob/4.1.2/jsonrpcserver/async_dispatcher.py#L91

metalg0su: I passes `methods` parameter to `dispatch` method.

The second positional argument `ws_methods` is the `methods` parameter.

loopchain/baseservice/node_subscriber.py
```diff
@@ -53,11 +51,12 @@ class UnregisteredException(Exception):
     pass
 
 
-def convert_response_to_dict(response: bytes) -> dict:
-    response_dict: dict = json.loads(response)
-    response_dict = _check_error_in_response(response_dict)
+def validate_response(response: str) -> str:
+    res: dict = json.loads(response)
+    res = _check_error_in_response(res)
```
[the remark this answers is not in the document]
metalg0su: I agree on your opinion.
I hope this will be treated in better way, in another PR soon. 

I think the PR can touch below:
 In latest version of `jsonrpcserver`, the library validates JSON-RPC schema in silent way. But some points in `loopchain` validates JSON-RPC schema again after the library does.

loopchain/baseservice/node_subscriber.py
yakkle: I thought that `global_methods` is usable in this case.

yakkle (approved):

loopchain/baseservice/node_subscriber.py
zsaladin: Keeping scope small is better in most cases if there is no reason to use it globally. 
Can you explain "can't replace `add`" in detail?

yakkle: I means below code in `__init__`.
```py
        ws_methods.add(self.node_ws_PublishHeartbeat)
        ws_methods.add(self.node_ws_PublishNewBlock)
```
So, It's not easy to remove `ws_methods` parameter.
And I already approved.

PR #497, opened 2020-04-01
[LC-924] Isolate legacy code.

windies21:
Add loopchain duplicate package as a legacy.

The 2.x version uses legacy packages.
3.x version will use loopchain package.


metalg0su:
Need to rebase this branch onto `develop3.0` branch

PR #506, opened 2020-04-10
[LC-937] InvokeResult and its pool

metalg0su:
# Goal
- Create `InvokeRequest`, `InvokeResult` and `InvokeResultPool`.

# R&R

## ...
- InvokeRequest: 
  - Wrapped data to be sent to ICON-Service.
  - Provides `serialize` to convert itself as dict.
- InvokeResult: 
  - Wrapped data of invoke result, returning as a result of invoke function.
- InvokeResultPool: 
  - Manages `InvokeResult`
  - Prunes expired (already committed) `InvokeResult`s.

### DIFF
--- a/loopchain/blockchain/invoke_result.py
+++ b/loopchain/blockchain/invoke_result.py
@@ -1,0 +1,238 @@
+from typing import TYPE_CHECKING, cast, Sequence, List, Union, Dict, OrderedDict, Optional
+
+from lft.consensus.messages.message import MessagePool, Message
+
+from legacy.channel.channel_property import ChannelProperty
+from legacy.utils.message_queue import StubCollection
+from loopchain.blockchain.blocks import BlockProver, BlockProverType, NextRepsChangeReason
+from loopchain.blockchain.blocks.v0_3 import BlockProver
+from loopchain.blockchain.transactions import TransactionSerializer, TransactionVersioner
+from loopchain.blockchain.types import Hash32, ExternalAddress
+
+if TYPE_CHECKING:
+    from loopchain.blockchain.votes.v1_0.vote import BlockVote
+
+
+class InvokeRequest:
+    def __init__(self, height: int,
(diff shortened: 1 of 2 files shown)


testcase/unittest/blockchain/test_invoke_result.py
```diff
@@ -0,0 +1,99 @@
...
+            signature=Signature.new(),
+            epoch=epoch_num,
+            round=round_num,
+            validators_hash=Hash32.new(),
+            next_validators_hash=Hash32.new(),
+            prev_votes_hash=Hash32.new(),
+            transactions_hash=Hash32.new(),
+            prev_state_hash=Hash32.new(),
+            prev_receipts_hash=Hash32.new(),
+            prev_logs_bloom=BloomFilter.new(),
```
metalg0su, line 78: Remove comma

testcase/unittest/blockchain/test_invoke_result.py
```diff
@@ -0,0 +1,99 @@
...
+        )
+        body = BlockBody(
+            transactions=["a"],
+            prev_votes="",
```
metalg0su, line 82: Remove comma

loopchain/blockchain/invoke_result.py
```diff
@@ -0,0 +1,238 @@
...
+            changed_reason: A reason why the prep list has been changed
+        """
+        self._id: bytes = f"{epoch_num}_{round_num}".encode()
+        self._epoch_num: int = epoch_num
+        self._round_num: int = round_num
+        self._added_transactions: Dict[str, dict] = added_transactions
+        self._reps_hash: Hash32 = reps_hash
+
+        self._next_preps: Optional[list] = None
+        self._next_preps_hash: Hash32 = Hash32.new()
```
windies21, line 129: In block 1.0, _next_preps_hash must be filled.(in this case with reps_hash) Block 1.0 does not use special blocks. That rule is only for 0.3 blocks.

PR #509, opened 2020-04-22
[LC-947] Remove dependency: `fluent-logger` (3.0)

metalg0su:
# Goal
- Apply #508 into `develop3.0`


metalg0su:
Continue to #522

PR #524, opened 2020-05-21
[LC-696] fix logical error that It can't sync same block height of leader node

noveljava:
Fix.
- Peers Can't sync same block height of leader node.


yakkle:

yakkle:
test this please

PR #527, opened 2020-05-25
[IS-1096] compatible dependency versions

yakkle:
**update requirements to compatible patch version**

   https://www.python.org/dev/peps/pep-0440/#version-specifiers


Makefile
```diff
@@ -41,7 +41,6 @@ all: install generate-key
 
 requires:
 	$(PIP_INSTALL) iconservice==1.6.1
-	$(PIP_INSTALL) iconcommons==1.1.2
```
metalg0su, line 44: I think you removed `iconcommons`, because it can be installed as a consequences of others' installation.

But `Loopchain` tries to import `iconcommons` in its code line, which makes me worried about removing it from the requirement.

Any opinion..?

yakkle: `iconcommons` will be install by `iconservice` or `iconrpcserver`.
It used only `launcher.py` and import with `iconservice` or `iconrpcserver` in launcher.
So, If you need `iconcommons`, `iconservice` or `iconrpcserver` also needed.

metalg0su: I see.
simpler is better.

Someday we can discuss the role of `iconcommons`.

metalg0su (approved):

hyerim-kim (approved):

windies21 (approved):

PR #501, opened 2020-04-02
[LC-928] Update ChannelStateMachine diagram

metalg0su:
# Goal
- Update ChannelStateMachine diagram

## AS-IS
image

## TO-BE
image

PR #535, opened 2020-06-01
[LC-928] Update ChannelStateMachine diagram (3.0)

metalg0su:
Apply #501 into develop3.0


metalg0su:
default (Jenkins) test failed due to mismatched version of `earlgrey`.
> Loopchain requires `0.0.4` and tries to install latest version (`>=0.0.4`)
> and `earlgrey` released `0.1.0`, which causes test failure.

It will be fixed after #527 is merged into `develop3.0`

PR #540, opened 2020-06-08
[IS-1056] manual rollback by command

yakkle:
- rollback behind to 1 block by run `loop -d -o loop_conf.json --rollback`

### DIFF
--- a/loopchain/channel/channel_service.py
+++ b/loopchain/channel/channel_service.py
@@ -39,7 +39,7 @@ from loopchain.utils.message_queue import StubCollection
 
 
 class ChannelService:
-    def __init__(self, channel_name, amqp_target, amqp_key):
+    def __init__(self, channel_name, amqp_target, amqp_key, rollback=False):
         self.__block_manager: BlockManager = None
         self.__score_container: CommonSubprocess = None
         self.__score_info: dict = None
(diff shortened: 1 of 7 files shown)


eunsoo-icon (approved):

windies21 (approved):

windies21 (approved):

loopchain/channel/channel_service.py
```diff
@@ -135,6 +138,37 @@ async def _serve():
...
+        """
+        await StubCollection().create_peer_stub()
+
+        results = await StubCollection().peer_stub.async_task().get_node_info_detail()
+        self._init_properties(**results)
+
+        self.__init_block_manager()
+        await self.__init_score_container()
+        await self.__inner_service.connect(conf.AMQP_CONNECTION_ATTEMPTS, conf.AMQP_RETRY_DELAY, exclusive=True)
+        await asyncio.sleep(0.01)   # sleep to complete peer service initialization
```
goldworm-icon, line 154: Isn't there an alternative except for sleeping?
0.01 sleeping is enough?

loopchain/channel/channel_service.py
```diff
@@ -135,6 +138,37 @@ async def _serve():
...
+
+    def _manual_rollback(self) -> str:
+        logging.debug("_manual_rollback() start manual rollback")
+        if self.block_manager.blockchain.block_height >= 0:
+            self.block_manager.rebuild_block()
+
+        if self.block_manager.request_rollback():
+            message = "rollback finished"
+        else:
+            message = "rollback cancelled"
```
goldworm-icon, line 167: Would it be better to add block height after rollback to message?
Is it possible to cancel rollback in progress?

yakkle: `loopchain` doesn't process rollback when failed rollback in `iconservice`.
In this case, show `rollback canceled` message.
It will show block height when rollback succeed.

PR #550, opened 2020-06-24
[LC-1008] Edit logic to build block without `rootHash` after `invoke`.

hyerim-kim:
Edit logic to build block without `rootHash` after `invoke`.

- Remove some duplicated code.
- Add some functions into BlockBuilder of version 0.3.

### DIFF
--- a/loopchain/blockchain/blockchain.py
+++ b/loopchain/blockchain/blockchain.py
@@ -591,8 +591,7 @@ class BlockChain:
             if not need_to_write_tx_info:
                 receipts = None
 
-            if next_prep and self.find_preps_addresses_by_roothash(
-                    Hash32.fromhex(next_prep['rootHash'], ignore_prefix=True)):
+            if next_prep and self.find_preps_addresses_by_roothash(block.header.next_reps_hash):
                 next_prep = None
 
             next_total_tx = self.__write_block_data(block, confirm_info, receipts, next_prep)
@@ -694,10 +693,10 @@ class BlockChain:
             self._write_tx(block, receipts, batch)
 
         if next_prep:
-            utils.logger.spam(
-                f"store next_prep in __write_block_data\nprep_hash({next_prep['rootHash']})"
-                f"\npreps({next_prep['preps']})")
-            self.write_preps(Hash32.fromhex(next_prep['rootHash'], ignore_prefix=True), next_prep['preps'], batch)
+            utils.logger.spam(f"store next_prep in __write_block_data\n"
+                              f"prep_hash({block.header.next_reps_hash})\n"
+                              f"preps({next_prep['preps']})")
+            self.write_preps(block.header.next_reps_hash, next_prep['preps'], batch)
 
         if confirm_info:
             if isinstance(confirm_info, list):
@@ -1164,7 +1163,7 @@ class BlockChain:
         self.__invoke_results[new_block.header.hash] = (tx_receipts, None)
         return new_block, tx_receipts
 
-    def _process_next_prep_legacy(self, _block: Block, block_builder: BlockBuilder, next_prep: dict):
+    def _process_next_prep_legacy(self, _block: Block, block_builder: BlockBuilder, reps: list, next_prep: dict):
         next_leader = _block.header.next_leader
 
         if next_prep:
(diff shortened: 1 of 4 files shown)


loopchain/blockchain/blockchain.py
```diff
@@ -1176,19 +1175,19 @@ def _process_next_prep_legacy(self, _block: Block, block_builder: BlockBuilder,
...
         block_builder.next_leader = next_leader
-        block_builder.reps = self.find_preps_addresses_by_header(_block.header)
+        block_builder.reps = reps
+        block_builder.next_reps = next_preps
         block_builder.next_reps_hash = next_preps_hash
 
-    def _process_next_prep(self, _block: Block, block_builder: BlockBuilder, next_prep: dict):
-        reps = self.find_preps_addresses_by_header(_block.header)
-
+    def _process_next_prep(self, _block: Block, block_builder: BlockBuilder, reps: list, next_prep: dict):
```
metalg0su, line 1190: > I'm afraid that this review is out of the PR's topic...

param `next_prep` could be `None`

I suggest:
```python
def _process_next_prep(..., next_prep: Union[dict]):
...
```

OR

```python
next_prep = response.get("prep", {}) # at 1285 line
```

loopchain/blockchain/blockchain.py
```diff
@@ -1164,7 +1163,7 @@ def genesis_invoke(self, block: Block, prev_block_=None) -> Tuple[Block, dict]:
         self.__invoke_results[new_block.header.hash] = (tx_receipts, None)
         return new_block, tx_receipts
 
-    def _process_next_prep_legacy(self, _block: Block, block_builder: BlockBuilder, next_prep: dict):
+    def _process_next_prep_legacy(self, _block: Block, block_builder: BlockBuilder, reps: list, next_prep: dict):
```
metalg0su, line 1166: See:

loopchain/blockchain/blocks/v0_3/block_builder.py
```diff
@@ -129,20 +130,27 @@ def _build_receipts_hash(self):
...
 
+    def build_next_reps_hash(self):
+        if self.next_reps_hash is not None:
+            return self.next_reps_hash
+
+        self.next_reps_hash = self._build_next_reps_hash()
+        return self.next_reps_hash
+
+    def _build_next_reps_hash(self):
+        block_prover = BlockProver((rep.extend() for rep in self.next_reps), BlockProverType.Rep)
```
metalg0su, line 151: How about checking `self.reps == self.next_reps` first, then use `self.reps_hash` instead, if two reps are identical?

I'm not sure but it might be faster in certain situations.

> get root hash of all leaves ALWAYS VS compare all items of sequence

hyerim-kim: I almost agree with you, most cases are the same maybe. But `reps` must be built earlier than `next_reps` always if do as your opinion. I worry about the dependency of the sequence between them. Let me know your opinion.

metalg0su: You're right.

My suggestion above is assured IF the reps hash is built BEFORE next reps hash.

Please keep it on.

metalg0su (approved):

loopchain/blockchain/blockchain.py
```diff
@@ -591,8 +591,7 @@ def __add_block(self, block: Block, confirm_info, need_to_write_tx_info=True, ne
             if not need_to_write_tx_info:
                 receipts = None
 
-            if next_prep and self.find_preps_addresses_by_roothash(
-                    Hash32.fromhex(next_prep['rootHash'], ignore_prefix=True)):
+            if next_prep and self.find_preps_addresses_by_roothash(block.header.next_reps_hash):
```
windies21, line 594: For block version 0.3. next_reps_hash is not always filled. See block specifications.  If you want the actual value of next_reps_hash, use reveal_next_reps_hash.

windies21, line 594: I think the way to compute the hash from next_prep itself is safer than using block's next_reps_hash.

loopchain/blockchain/blockchain.py
```diff
@@ -694,10 +693,10 @@ def __write_block_data(self, block: Block, confirm_info, receipts, next_prep):
             self._write_tx(block, receipts, batch)
 
         if next_prep:
-            utils.logger.spam(
-                f"store next_prep in __write_block_data\nprep_hash({next_prep['rootHash']})"
-                f"\npreps({next_prep['preps']})")
-            self.write_preps(Hash32.fromhex(next_prep['rootHash'], ignore_prefix=True), next_prep['preps'], batch)
+            utils.logger.spam(f"store next_prep in __write_block_data\n"
```
windies21, line 696: My suggestion is calculate and use the reps_hash in this method.

loopchain/blockchain/blocks/v0_4/block_builder.py
```diff
@@ -1,12 +1,8 @@
-from typing import List
-
-from loopchain.blockchain.blocks import BlockBuilder as BaseBlockBuilder, BlockProverType
 from loopchain.blockchain.blocks.block import NextRepsChangeReason
 from loopchain.blockchain.blocks.v0_3 import BlockBuilder
```
windies21, line 2: How about this:
from loopchain.blockchain.blocks.v0_3 import BlockBuilder as BaseBlockBuilder

loopchain/blockchain/blockchain.py
```diff
@@ -1176,19 +1169,19 @@ def _process_next_prep_legacy(self, _block: Block, block_builder: BlockBuilder,
             if change_reason == NextRepsChangeReason.TermEnd:
                 next_leader = ExternalAddress.empty()
 
-            next_preps_hash = Hash32.fromhex(next_prep["rootHash"], ignore_prefix=True)
+            next_preps = [ExternalAddress.fromhex(prep["id"]) for prep in next_prep["preps"]]
+            next_preps_hash = None  # to rebuild next_reps_hash
         else:
             # P-Rep list has no changes
-            next_leader = _block.header.next_leader
+            next_preps = reps
```
windies21, line 1176: Does it fit the v0.3 block specification? see. https://icon-project.atlassian.net/wiki/spaces/LO/pages/92045385/Block+0.3

windies21 (approved):

PR #552, opened 2020-07-02
[LC-1016] Prune InvokeResult after adding block

metalg0su:
# Goal
- Prune InvokePool messages (results of IS invoke) right after writing Block in DB.


yakkle:
stale

PR #555, opened 2020-07-03
[IS-1136] rename testcase/unittest directory

yakkle:
- fixed to `tests/unit` because of conflict with python `unittest` module


metalg0su:
You changed some lines with `test_util` as relative import while other lines kept as it is.

---
Two patterns found:
```python
from tests.unit.blockchain.conftest import TxFactory, TxBuilderFactory
```
OR 
```python
from ..unit import test_util
```

metalg0su (approved):

yakkle:
> You changed some lines with `test_util` as relative import while other lines kept as it is.
> Do you have any ideas on this?
> 

I don't know what is better. I just leave link for reference.

https://stackoverflow.com/a/38231766/7155818
https://docs.pytest.org/en/stable/goodpractices.html

PR #557, opened 2020-07-14
[LC-1024] Exit BlockSync when the current block is 1.0

metalg0su:
# Goal
- In BlockSync, no more request/sync blocks above 1.0 (1.0 height include)


noveljava:
Looks Good To Me.

metalg0su:
This is out of our milestone scope.
Close first then reopen when needed.

Thank you.

PR #567, opened 2020-07-31
[LC-1038] Fix broken CI

metalg0su:
# Goal
- Fix broken tests.

> This problem may be caused from other dependencies.
> Patch first into develop3.0


metalg0su:
Merging into develop is suspended.

Merge into develop3.0 first.

PR #568, opened 2020-07-31
[LC-1039] Fix a bug when `url_input` is `None`.

hyerim-kim:
When `url_input` is `None`, the exception occurs `TypeError`. Because Nonetype is not iterable.


windies21 (approved):

metalg0su:
Pushed simple commit contains test case.
Consider to use it.

PR #573, opened 2020-08-06
[LC-1042] Update last_unconfirmed_block on LFT consensus

metalg0su:
# Goal
- `BlockHeightRequest` now triggers to return highest block height, contains `last_unconfirmed_block`
- `BlockRequest`now triggers to return highest block, contains `last_unconfirmed_block`
- Update BlockChain.last_unconfirmed_block in receiving or broadcasting any block.

# Why `last_unconfirmed_block` needed?
- Nodes need to exchange informations each other, not only known (committed) info, but unknown (needs to be voted) info!

### DIFF
--- a/loopchain/blockchain/blockchain.py
+++ b/loopchain/blockchain/blockchain.py
@@ -1443,7 +1443,10 @@ class BlockChain:
 
         return block_builder
 
-    def try_update_last_unconfirmed_block(self, unconfirmed_block: 'Block'):
+    def try_update_last_unconfirmed_block(self, unconfirmed_block: 'Block') -> bool:
         """Note that the method expects an input as Block 1.0+"""
         if not self.__last_block or unconfirmed_block.header.height == self.__last_block.header.height + 2:
             self.last_unconfirmed_block = unconfirmed_block
+            return True
+
+        return False
--- a/loopchain/blockchain/blockchain.py
+++ b/loopchain/blockchain/blockchain.py
@@ -460,7 +460,7 @@ class BlockChain:
         """
         return self.__find_block_by_key(block_hash.hex().encode(encoding='UTF-8'))
 
-    def find_block_by_height(self, block_height):
+    def find_block_by_height(self, block_height) -> Optional['Block']:
         """find block in DB by its height
 
         :param block_height: int,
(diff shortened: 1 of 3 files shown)


loopchain/channel/channel_inner_service.py
```diff
@@ -637,7 +637,9 @@ async def block_request(self, request_from: str, height: int):
         Note that BlockRequest allows only Block 1.0+
         """
         blockchain = self._channel_service.block_manager.blockchain
-        block: Block_V1_0 = blockchain.find_block_by_height(height)
+        block: Optional[Block_V1_0] = blockchain.find_block_by_height(height)
+        if not block:
+            return  # FIXME: Ignore response? or Send client to failure message?
```
metalg0su, line 642: Please comment on this.

loopchain/blockchain/blockchain.py
```diff
@@ -1442,3 +1442,8 @@ def makeup_block(self,
             self.__add_tx_to_block(block_builder)
 
         return block_builder
+
+    def try_update_last_unconfirmed_block(self, unconfirmed_block: 'Block'):
+        """Note that the method expects an input as Block 1.0+"""
+        if not self.__last_block or unconfirmed_block.header.height == self.__last_block.header.height + 2:
```
noveljava, line 1448: (self.__last_block.header.height + 1) also will need to update.

windies21, line 1448: If this condition check is to determine a normal leader, shouldn't the effect of this check be reflected in send_unconfirmed_block(of runner.py) as well? If not, wouldn't it be okay to not have this condition check?

metalg0su: In normal case, no need to request `last_block.header.height + 1` block, because it is candidate block, which makes evidence of `last_block`.

But suppose that there is a node whose last_block height is `last_block.header.height - 1`.
He or she can request `last_block.header.height + 1`.

We will discuss again soon when we deal with sync problems.

loopchain/channel/channel_inner_service.py
windies21: I think it needs CommonReply with Fail code.

windies21: I mean we have to change BlockRequest of peer_outer to treat in this case.

windies21: # FIXME -> # This case have to handle by the requesting node.

metalg0su: Fixed.

PR #577, opened 2020-08-11
[LC-1050] Modulate round start time

metalg0su:
# Goal
- if the network commits block too fast, it should be slower
- if the network commits block too slow, it should be faster.


yakkle:
stale

PR #579, opened 2020-08-11
[LC-1047] Use only valid votes when leader creates block

metalg0su:


loopchain/blockchain/blocks/v1_0/block_factory.py
```diff
@@ -73,7 +74,10 @@ async def create_data(self, data_number: int, prev_id: bytes, epoch_num: int, ro
         block_builder.validators_hash = pre_invoke_response.validators_hash
 
         if prev_votes:
-            prev_vote = prev_votes[0]  # FIXME
+            try:
+                prev_vote = next(vote for vote in prev_votes if vote.is_real())
+            except StopIteration:
+                raise RuntimeError("Something wrong...") from StopIteration
```
metalg0su, line 80: Need useful msg and exc type.
Please comment on this.

windies21: If the prev_votes are checked then this exception cannot be reached.

windies21: So.... Just change the message.... How about this?
RuntimeError ("Unverified prev_votes.")

metalg0su: Applied!

windies21 (approved):

PR #581, opened 2020-08-21
[LC-988] Correct epoch number when round starts

metalg0su:
(Resolve WIP someday after all related-PR are merged.)

# Goal
- Fix epoch num when round starts

> as-is: non-increasing static value `1`
> to-do: increment when validators are changed


yakkle:
stale

PR #582, opened 2020-09-02
[LC-1058] Remove IS_BROADCAST_ASYNC option and related code(unreached).

windies21:
Remove IS_BROADCAST_ASYNC option and related code(unreached).


metalg0su (approved):
Good

PR #628, opened 2021-04-05
[LC-1072] Update the reject_create_tx properties to the subprocess according to the following config values (TX_COUNT_TO_START_REJECT and TX_COUNT_TO_RESUME_ACCEPT).

windies21:


loopchain/channel/channel_inner_service.py
```diff
@@ -429,12 +434,17 @@ def crash_callback_in_join_thread(process: ModuleProcess):
         logging.info(f"Channel({ChannelProperty().name}) TX Receiver: initialized")
 
     def update_sub_services_properties(self, **properties):
-        logging.info(f"properties {properties}")
-        stub = StubCollection().channel_tx_creator_stubs[ChannelProperty().name]
-        asyncio.run_coroutine_threadsafe(stub.async_task().update_properties(properties), self.__loop_for_sub_services)
+        logging.debug(f"properties {properties}")
```
leeheonseung, line 437: Is there any reason I need to handle both logic in the update_sub_services_properties function?

I think there is a problem with the checklogic below.
`if len(properties) == 1 and "reject_create_tx" in properties:
            target_stubs.pop('tx_receiver')`

loopchain/peer/block_manager.py
```diff
@@ -648,6 +648,10 @@ def vote_unconfirmed_block(self, block: Block, round_: int, is_validated):
             reps_hash=target_reps_hash
         )
 
+        if (self.__channel_service.inner_service.get_sub_services_properties('reject_create_tx') and
```
leeheonseung, line 651: The logic below doesn't work for citizen nodes.
CitizenNode also needs to be fixed so that there are no problems.

loopchain/configure_default.py
```diff
@@ -126,6 +126,8 @@ class ConsensusAlgorithm(IntEnum):
 # The total size of the transactions in a block.
 MAX_TX_SIZE_IN_BLOCK = 1 * 1024 * 1024  # 1 MB is better than 2 MB (because tx invoke need CPU time)
 MAX_TX_COUNT_IN_ADDTX_LIST = 128  # AddTxList can send multiple tx in one message.
+TX_COUNT_TO_START_REJECT = 10000  # Maximum allowed size of unconfirmed tx pools.
```
leeheonseung, line 129: Is there a reason for using two constants instead of one constant?

loopchain/channel/channel_inner_service.py
```diff
@@ -974,8 +978,24 @@ def init_sub_services(self):
             raise Exception("Must call this function in thread of self.loop")
         self._task.init_sub_service(self.loop)
 
+    def need_to_update_properties(self, **properties) -> bool:
```
leeheonseung, line 981: Is there any reason not to use dictionary equal? It compares only 1 depth.

loopchain/peer/block_manager.py
```diff
@@ -648,6 +648,10 @@ def vote_unconfirmed_block(self, block: Block, round_: int, is_validated):
             reps_hash=target_reps_hash
         )
 
+        if (self.__channel_service.inner_service.need_to_update_properties(reject_create_tx=None)
```
leeheonseung, line 651: I would like to be able to do it as a standalone function by adding a timer, since additional functions may be needed in the future as well.

leeheonseung (approved):

hyerim-kim (approved):

